PR #3, opened 2020-01-16
feat(WEBRTC-53): migration verto to telnyxRTC

DeividVeloso:
- Add support to modules verto and telnyx_rtc
 - Rename public interfaces from Verto to TelnyxRTC

## 📝 To Do

- [x] All linters pass
- [x] All tests pass

## ✋ Manual testing

1. Access http://localhost:6006/?path=/story/webdialer--example
2. Try to make a call

### Desktop

- [ ] Edge (latest)
- [x] Chrome
- [ ] Firefox
- [ ] Safari


src/TelnyxRTCClient/TelnyxRTCClient.test.ts
```diff
@@ -23,8 +23,8 @@ Object.defineProperty((global as any).navigator, 'mediaDevices', {
   configurable: true,
 });
 
-describe('VertoClient', () => {
-  const client = new VertoClient({
+describe('TelnyxRTCClient', () => {
```
SuaYoo: Could you add a spec for the new `module` parameter?

src/TelnyxRTC/Verto.ts
```diff
@@ -130,6 +132,7 @@ export default class Verto {
     this.dialog = null;
     this.dialogs = {};
     this.params = params;
+    this.module = params.module;
```
SuaYoo: Can you throw an error early here if `module` is not `verto` or `telnyx_rtc`? That way we don't force the caller to wait until a method is called to throw a type error.

DeividVeloso: I will do that

src/TelnyxRTCClient/TelnyxRTCClient.test.ts
DeividVeloso: Done!

PR #32, opened 2020-07-30
fix(WEBRTC-124): fixed audio call on Firefox browser

DeividVeloso:
- Fixed audio call problem on Firefox due to the **createOffer** method being called with both options of MediaStream enabled
`{ offerToReceiveAudio: true, offerToReceiveVideo: true }` even though I only use an **audio** options.

In the future, when we support `unified-plan` we will need to change this 
```
{ offerToReceiveAudio: true, offerToReceiveVideo: true }
```

To: 
```
 addTransceiver('audio')
```

These options are legacy
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createOffer

## 📝 To Do

- [x] All linters pass
- [x] All tests pass

## ✋ Manual testing

1. In root folder `npm i ` && `npm run build`
2. Navigate into `examples/react`
3. Run `npm i` && `npm run setup` && `npm run start`
4. Open the example in a Firefox browser and try to make a call and answer

### Desktop

- [ ] Edge (latest)
- [x] Chrome
- [x] Firefox
- [x] Safari

### DIFF
--- a/src/Modules/Verto/index.ts
+++ b/src/Modules/Verto/index.ts
@@ -8,6 +8,7 @@ import { trigger } from './services/Handler';
 import { localStorage } from './util/storage';
 import VertoHandler from './webrtc/VertoHandler';
 import { isValidOptions } from './util/helpers';
+import webrtcAdapter from 'webrtc-adapter';
 
 export const VERTO_PROTOCOL = 'verto-protocol';
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -8892,9 +8892,9 @@
       "dev": true
     },
     "webrtc-adapter": {
-      "version": "7.6.3",
-      "resolved": "https://registry.npmjs.org/webrtc-adapter/-/webrtc-adapter-7.6.3.tgz",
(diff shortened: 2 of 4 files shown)


src/Modules/Verto/index.ts
```diff
@@ -8,6 +8,7 @@ import { trigger } from './services/Handler';
 import { localStorage } from './util/storage';
 import VertoHandler from './webrtc/VertoHandler';
 import { isValidOptions } from './util/helpers';
+import webrtcAdapter from 'webrtc-adapter';
```
hugobdev, line 11: is this unused?

DeividVeloso: It works like a polyfill https://www.npmjs.com/package/webrtc-adapter

SuaYoo: It may be more obvious to just do `import 'webrtc-adapter';` in that case, or to log `webrtcAdapter .browserDetails.version` if a DEBUG flag is set

PR #53, opened 2020-11-09
docs: Add TelnyxRTC constructor docs

hugobdev:
Overwrote the constructor in the TelnyxRTC class because the docs talk
about the creation of the TelnyxRTC class, instead of any of the
inhereted classes. It would be wrong to document BrowserSessions
constructor as creating a TelnyxRTC instance.

### DIFF
--- a/packages/js/src/TelnyxRTC.ts
+++ b/packages/js/src/TelnyxRTC.ts
@@ -7,7 +7,7 @@ export default class TelnyxRTC extends TelnyxRTCClient {
    *
    * @param options An object with options.
    * @param options.login_token The JSON Web Token (JWT) to authenticate with your SIP Connection. This is the recommended authentication strategy. [See how to create one](https://developers.telnyx.com/docs/v2/webrtc/quickstart).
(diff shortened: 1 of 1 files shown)


DeividVeloso (approved):

packages/js/src/TelnyxRTC.ts
```diff
@@ -1,3 +1,47 @@
 import TelnyxRTCClient from './Modules/Verto';
+import { ITelnyxRTCOptions } from './Modules/Verto/util/interfaces';
 
-export default class TelnyxRTC extends TelnyxRTCClient {}
+export default class TelnyxRTC extends TelnyxRTCClient {
+  /**
+   * Creates a new `TelnyxRTC` instance with the provided options.
+   *
+   * @param options An object with options.
+   * @param options.login_token The JSON Web Token (JWT) to authenticate with your SIP Connection. This is the recommended authentication strategy. [See how to create one](https://developers.telnyx.com/docs/v2/webrtc/quickstart).
```
SuaYoo, line 9: Nice, maybe verify and mention that `login_token` will take precedence over `login`/`password`?

PR #55, opened 2020-11-10
Add React client context provider and hooks

SuaYoo:
Adds additional TelnyxRTC abstractions via hooks and Context.

## 📝 To Do

- [x] All linters pass
- [x] All tests pass
- [x] Change documentation based on my changes

## ✋ Manual testing

1. Run `npm start`
2. [Generate a token](https://developers.telnyx.com/docs/v2/webrtc/quickstart) and save as `REACT_APP_TELNYX_LOGIN_TOKEN` in a `.env` file.
3. Open new shell and `cd example`
4. Start example app with `npm start`. Verify localhost:3000 matches screenshot (below)

### Desktop

- [ ] Edge (latest)
- [ ] Chrome
- [x] Firefox
- [ ] Safari

## 📸 Screenshots

Screen Shot 2020-11-10 at 4 36 16 PM


packages/react-client/README.md
```diff
@@ -12,19 +12,89 @@ npm install --save @telnyx/react-client
...
+    <div>
+      {activeCall &&
+        activeCall.state === 'ringing' &&
+        'You have an incoming call.'}
+    </div>
+  );
+}
+```
+
+### with `useEvents`
```
SuaYoo, line 50: Maybe `useCallbacks` is a better name?

packages/react-client/src/useTelnyxClient.ts
```diff
@@ -1,52 +1,34 @@
...
- * const { client, call, clientState } = useTelnyxClient({ login_token })
+ * const client = useTelnyxClient({ login_token })
  *
  * // Or, login using your SIP Connection username and password
- * // const { client, clientState } = useTelnyxClient({ login, password })
+ * // const client = useTelnyxClient({ login, password })
+ *
+ * client.on('telnyx.notification', ({ call }) => {
+ *   console.log(call)
+ * })
```
hugobdev, line 31: If you add this between backticks, VS Code will highlight the code just like Markdown

packages/react-client/src/useTelnyxClient.ts
```diff
@@ -55,103 +37,38 @@ type CredentialOptions = TokenCredentialOptions | UsernameCredentialOptions;
...
-
       telnyxClientRef.current?.disconnect();
-      telnyxClientRef.current = undefined;
     };
-  }, [credentialParam]);
+  }, []);
 
-  return { client: telnyxClientRef.current, call, clientState };
+  return telnyxClientRef.current;
 }
```
hugobdev, line 72: To make sure all this code is only running in the browser, we should run it all inside `useEffect` I believe.

hugobdev: If you instantiate inside the `useEffect` you would have to return the ref from the function, not the `.current`

packages/react-client/README.md
```diff
@@ -12,19 +12,89 @@ npm install --save @telnyx/react-client
 
 ## Usage
 
-```tsx
-import React, { Component } from 'react';
+```jsx
+// App.jsx
+import { TelnyxClientProvider } from '@telnyx/react-client';
```
hugobdev, line 17: We still call the main package's class `TelnyxRTC`, so I think we should got with `TelnyxRTCProvider`, `useTelnyxRTC`, and so on.

packages/react-client/src/useTelnyxClient.ts
SuaYoo: @hugobdev I don't think `TelnyxRTC` needs access to the DOM any more since we no longer provider an audio or video element ID during client instantiation. Running this in a hook would also require all consumers to check for the existence of the client, which I don't think necessary. Did you have another concern in mind?

hugobdev (approved):
This is going to be a game changer for react integration

PR #67, opened 2020-11-19
Document release instructions & add release-it to react-client

SuaYoo:
Documents release/publish instructions for `@telnyx/webrtc` and `@telnyx/react-client`

I published react-client manually as a prerelease: https://github.com/team-telnyx/webrtc/releases/tag/react-client%2Fv1.0.0-beta.0

Note: I haven't figured out how to get around the patch version bump (yet) even when there are no code changes. See https://github.com/conventional-changelog/conventional-changelog/issues/399 and https://github.com/conventional-changelog/standard-version/issues/681


.github/workflows/js-draft-release.yml
```diff
@@ -7,11 +7,6 @@ on:
     paths:
     - 'packages/js/src/**'
   workflow_dispatch:
-    inputs:
```
SuaYoo, line 10: Removed this input since the GH interface differentiates manual triggers easily

hugobdev (approved):
What should our PR merge strategy be if we are doing conventionalcommits tied to releases? I think it would be great to force that all commits that go to main are following the convention.

DeividVeloso (approved):

PR #75, opened 2020-12-02
feat(WEBRTC-232): remove refreshToken and blade methods

DeividVeloso:
We don't support `refreshToken` methods, so I removed these methods related to JWT expiration.

## 📝 To Do

- [x] All linters pass
- [x] All tests pass

## ✋ Manual testing

1. Navigate into `package/js` && npm run build && `npm run test`
2. Run `npm pack` get the `.tgz` file and install inside `examples/react-video` using `npm i`
3. Make a call and receive a call to see if it is working.

### Desktop

- [ ] Edge (latest)
- [x] Chrome
- [ ] Firefox
- [ ] Safari

### DIFF
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -2,7 +2,6 @@ import * as log from 'loglevel';
 import { v4 as uuidv4 } from 'uuid';
 import logger from './util/logger';
 import Connection from './services/Connection';
-import Setup from './services/Setup';
 import BaseMessage from './messages/BaseMessage';
 import {
   deRegister,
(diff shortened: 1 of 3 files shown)


packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -257,36 +226,7 @@ export default abstract class BaseSession {
...
-      this.nodeid = nodeid;
-      this.master_nodeid = master_nodeid;
-      this._emptyExecuteQueues();
-      this._pong = null;
-      this._keepAlive();
-      trigger(SwEvent.Ready, this, this.uuid);
-      logger.info('Session Ready!');
-    }
-  }
+  protected async _onSocketOpen() {}
```
hugobdev, line 229: Are you going to remove these here as well?

hugobdev (approved):
This doesn't make sense with the way we implemented tokens, so good you removed. Thank you.

packages/js/src/Modules/Verto/BaseSession.ts
DeividVeloso: No, I am not. This one is used to register the SocketOpen that comes from his child Verto.Ts

packages/js/src/Modules/Verto/index.ts
```diff
@@ -64,6 +65,7 @@ export default class Verto extends BrowserSession {
       this.sessionid = response.sessid;
       sessionStorage.setItem(SESSION_ID, this.sessionid);
       trigger(SwEvent.Ready, this, this.uuid);
+      logger.info('Session Ready!');
```
SuaYoo, line 68: What is the purpose of this log?

DeividVeloso: I copied from the method `_onSocketOpen` that `Blade` was using, I think it is helpful when you are debugging the service on production.

DeividVeloso: You can check here https://github.com/team-telnyx/webrtc/pull/75/files#diff-68dd50822bd145df05bdf53bbc4b6e2242cafe1e4d03aa94f2a886babcc1400dL287

SuaYoo (approved):

PR #81, opened 2020-12-03
Document event types

SuaYoo:
Documents possible event names in `on` method.

Note: I did some research into ways to document event names in a similar fashion but didn't find a solution that rendered nicely with typedoc. I've tried documenting the enum directly, but wasn't getting a typedoc output at all. Open to other ideas.

## 📸 Screenshots

Ran `npm run docs` and previewed the output:
Screen Shot 2020-12-02 at 5 26 17 PM

### DIFF
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -190,20 +190,6 @@ export default abstract class BaseSession {
   /**
    * Attaches an event handler for a specific type of event.
    *
-   * ## Examples
-   *
-   * Subscribe to the `telnyx.ready` and `telnyx.error` events.
-   *
-   * ```js
-   * const client = new TelnyxRTC(options);
-   *
-   * client.on('telnyx.ready', (client) => {
-   *   // Your client is ready!
-   * }).on('telnyx.error', (error) => {
-   *   // Got an error...
-   * })
-   * ```
-   *
    * ## Events
    * |   |   |
    * |---|---|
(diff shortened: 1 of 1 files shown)


packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -206,6 +201,22 @@ export default abstract class BaseSession {
...
+   * | `telnyx.socket.close` | The WebSocket connection is set to close |
+   * | `telnyx.socket.error` | An error occurred at the WebSocket level |
+   * | `telnyx.socket.message` | The client has received a message through WebSockets |
+   *
+   * @param eventName Event name.
+   * @param callback Function to call when the event comes.
+   *
+   * @return The client object itself.
    */
   on(eventName: string, callback: Function) {
```
hugobdev, line 221: We could type eventName as an enum. I guess we would still have to document what each means though.

Having it as an Enum here will provide a nicer developer experience, as you get autocomplete when providing the string or can use `Enum.Event` kind of syntax.

packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -188,11 +188,6 @@ export default abstract class BaseSession {
   /**
    * Attaches an event handler for a specific type of event.
    *
-   * @param eventName Event name.
-   * @param callback Function to call when the event comes.
-   *
-   * @return The client object itself.
-   *
    * ## Examples
```
hugobdev, line 191: You could change this to be `@examples` so that you don't have to provide than on top of `@param`. You will still need to have `@param` after `## Events` though.

hugobdev (approved):

packages/js/src/Modules/Verto/BaseSession.ts
SuaYoo: I don't think this is possible with string `Enum` values, but will investigate

PR #84, opened 2020-12-03
Fix changelog versioning

SuaYoo:
Prevents creating a git tag and bumping package.json on merge to `main`, so that we can merge multiple times without generating multiple versions.


SuaYoo:
Merging without review to test this out

PR #107, opened 2020-12-12
Streamline JS README

SuaYoo:
Streamlines the @telnyx/webrtc README by cross-linking with examples in docs, to reduce the amount of duplicated (and erroneous) example code.

Side by side comparison: [Current README](https://github.com/team-telnyx/webrtc/blob/main/packages/js/README.md) | [Proposed README](https://github.com/team-telnyx/webrtc/blob/Update_js_readme/packages/js/README.md)

PR #108, opened 2020-12-12
Hide internal docs

SuaYoo:
Hides internal docs by explicitly specifying public docs. I started out with `TelnyxRTC` and `Call` based on these proposed README changes: https://github.com/team-telnyx/webrtc/pull/107

To have a declaration appear in docs, simply comment the module. I've also added an `@apidocs` tag just so we can even more explicitly keep track of which modules to document. Scroll near the bottom of the "Files changed" to view the TS file changes.

[Browse proposed docs here](https://github.com/team-telnyx/webrtc/tree/WEBRTC-291_Hide_internal_docs/packages/js/docs/ts)

TODO: I haven't figured out a way to hide the "_Inherited from_" link yet. May need to hide this in the final template.


DeividVeloso (approved):
Nice

PR #116, opened 2020-12-16
Update exported interfaces

SuaYoo:
Updates TS interfaces exported from `@telnyx/webrtc` to match the TS docs.

See TODOs, we should consolidate these with `CallOptions` and `Call` at some point, or release a breaking change and remove the interface exports.


packages/js/src/Modules/Verto/util/interfaces.ts
```diff
@@ -105,7 +105,7 @@ export interface IVideoSettings extends MediaTrackConstraints {
   camLabel?: string;
 }
 
-export interface ICall {
+export interface IVertoCall {
```
SuaYoo, line 108: Renamed these because I was confused at the multiple `ICall` interfaces

DeividVeloso (approved):
Nice.

PR #145, opened 2021-02-08
fix(WEBRTC-362): add delay when call stop play audio, because it should call pause method after play method

DeividVeloso:
- add delay when call stop playing audio, because it should call pause method after play method

## 📝 To Do

- [x] All linters pass
- [ ] All tests pass

## ✋ Manual testing

1. Access any WebRTC project and add the "auto-answer" method
```
    if (notification.call.state === 'ringing') {
            notification.call.answer();
   }
```
2. open two instances of WebRTC app and make a call to each other
3. You shouldn't get any errors on the console and you should hear the audio playing and stop the audio.

### Desktop

- [ ] Edge (latest)
- [x] Chrome
- [ ] Firefox
- [ ] Safari

### Error
image

### DIFF
--- a/packages/js/src/Modules/Verto/webrtc/helpers.ts
+++ b/packages/js/src/Modules/Verto/webrtc/helpers.ts
@@ -6,6 +6,7 @@ import {
   IVertoCallOptions,
   IWebRTCSupportedBrowser,
   IWebRTCInfo,
+  IAudio,
 } from './interfaces';
 
 const getUserMedia = async (
@@ -609,12 +610,19 @@ function getWebRTCSupportedBrowserList(): Array<IWebRTCSupportedBrowser> {
   ];
 }
 
-function createAudio(file, id): HTMLAudioElement | null {
+function createAudio(file, id): IAudio | null {
+  const elementExist = document.getElementById(id) as IAudio;
+
+  if (elementExist) {
+    return elementExist;
+  }
+
   if (file && id) {
-    const ringAudio = document.createElement('audio');
+    const ringAudio = document.createElement('audio') as IAudio;
     ringAudio.id = id;
     ringAudio.loop = true;
     ringAudio.src = file;
+    ringAudio.preload = 'auto';
     ringAudio.load();
     document.body.appendChild(ringAudio);
     return ringAudio;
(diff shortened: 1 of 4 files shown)


packages/js/src/Modules/Verto/webrtc/helpers.ts
```diff
@@ -563,14 +593,59 @@ function getWebRTCSupportedBrowserList(): Array<IWebRTCSupportedBrowser> {
...
+        },
       ],
     },
   ];
 }
 
+function createAudio(file, id): HTMLAudioElement | null {
+  if (file && id) {
+    const ringAudio = document.createElement('audio');
+    ringAudio.id = id;
```
SuaYoo, line 615: We should to check if an element with this ID exists first, in case the client is initialized multiple times.

packages/js/src/Modules/Verto/webrtc/helpers.ts
```diff
@@ -563,14 +593,59 @@ function getWebRTCSupportedBrowserList(): Array<IWebRTCSupportedBrowser> {
...
+function stopAudio(audioElement: HTMLAudioElement): void {
+  // Add delay to wait until play() returns the promise
+  // https://developers.google.com/web/updates/2017/06/play-request-was-interrupted
+  setTimeout(() => {
+    if (audioElement) {
+      audioElement.pause();
+      audioElement.currentTime = 0;
+    }
+  }, 1000);
+}
```
SuaYoo, line 647: Setting a timeout could be really hard to debug, especially since the 1 second seems fairly arbitrary--if the ringtone audio file is large, it may take longer to load, or if its small, it could load much quicker.

I would store the promise somehow, maybe like this (untested):
```suggestion
function playAudio(audioElement: HTMLAudioElement): void {
  if (audioElement) {
    audioElement._playFulfilled = false;
    audioElement._promise = audioElement
      .play();
      
    audioElement._promise.then(() => {
        audioElement._playFulfilled = true;
      })
      .catch((error) => {
        console.error('playAudio', error);
        
        audioElement._playFulfilled = true;
      });
  }
}

function stopAudio(audioElement: HTMLAudioElement): void {
  if (!audioElement) return;
  // Add delay to wait until play() returns the promise
  // https://developers.google.com/web/updates/2017/06/play-request-was-interrupted
  if (audioElement._playFulfilled) {
      audioElement.pause();
      audioElement.currentTime = 0;
  } else if (audioElement._promise && audioElement._promise.then) {
    audioElement._promise.then(() => {
      audioElement.pause();
      audioElement.currentTime = 0;
    })
  } else {
    setTimeout(() => {
      if (audioElement) {
        audioElement.pause();
        audioElement.currentTime = 0;
      }
    }, 1000);
  }
}
```

packages/js/src/Modules/Verto/webrtc/helpers.ts
```diff
@@ -563,14 +593,59 @@ function getWebRTCSupportedBrowserList(): Array<IWebRTCSupportedBrowser> {
...
     },
   ];
 }
 
+function createAudio(file, id): HTMLAudioElement | null {
+  if (file && id) {
+    const ringAudio = document.createElement('audio');
+    ringAudio.id = id;
+    ringAudio.loop = true;
+    ringAudio.src = file;
```
SuaYoo, line 617: Preloading the audio should also help remove the delay before pause:
```suggestion
    ringAudio.src = file;
    ringAudio.preload = "auto";
```

SuaYoo (changes requested):

packages/js/src/Modules/Verto/webrtc/helpers.ts
DeividVeloso: Done! It worked very well.

SuaYoo (approved):
Nice one!

PR #235, opened 2021-12-09
fix(ENGDESK-12851): add code to avoid multiples REGED messages on the client side

DeividVeloso:
- add code to avoid multiples REGED messages on the client-side

## 📝 To Do

- [x] All linters pass
- [x] All tests pass

## ✋ Manual testing

1. You can run any example project inside `packages/js/examples`
2. Make a npm link
3. Register in the SDK and see if the `telnyx.ready` will dispatch only once.

### Desktop

- [ ] Edge (latest)
- [x] Chrome
- [ ] Firefox
- [ ] Safari

## 📸 Screenshots

| Description | Screenshot |
| ----------- | ---------- |
| Desktop     |image (12)|

### DIFF
--- a/packages/js/src/Modules/Verto/util/helpers.ts
+++ b/packages/js/src/Modules/Verto/util/helpers.ts
@@ -133,15 +133,18 @@ export const isValidOptions = ({
 };
 
 export const getGatewayState = (msg: IMessageRPC): GatewayStateType | '' => {
-  const hasStateResult: GatewayStateType | '' =
-    msg && msg.result && msg.result.params && msg.result.params.state
-      ? msg.result.params.state
-      : '';
+  let stateResult: GatewayStateType | '' = '';
+  let stateParam: GatewayStateType | '' = '';
 
-  const hasStateParam: GatewayStateType | '' =
-    msg && msg.params && msg.params.state ? msg.params.state : '';
+  if (msg?.result?.params?.state) {
+    stateResult = msg?.result?.params?.state;
+  }
 
-  const gateWayState = hasStateResult || hasStateParam;
+  if (msg?.params?.state) {
+    stateParam = msg?.params?.state;
+  }
+
+  const gateWayState = stateResult || stateParam;
 
   return gateWayState;
 };
--- a/packages/js/src/Modules/Verto/util/helpers.ts
+++ b/packages/js/src/Modules/Verto/util/helpers.ts
@@ -1,6 +1,7 @@
-import { IVertoOptions } from './interfaces';
+import { IMessageRPC, IVertoOptions } from './interfaces';
 import logger from './logger';
 import { STORAGE_PREFIX } from './constants';
+import { GatewayStateType } from '../webrtc/constants';
 
 // hack to remove undefined values from the object
 export const deepCopy = (obj: Object) => JSON.parse(JSON.stringify(obj));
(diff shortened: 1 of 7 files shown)


packages/js/src/Modules/Verto/util/helpers.ts
```diff
@@ -130,3 +131,17 @@ export const isValidOptions = ({
   const isToken = login_token;
   return Boolean(isLogin || isToken);
 };
+
+export const getGatewayState = (msg: IMessageRPC): GatewayStateType | '' => {
+  const hasStateResult: GatewayStateType | '' =
+    msg && msg.result && msg.result.params && msg.result.params.state
+      ? msg.result.params.state
+      : '';
```
arghav, line 139: This statement is quite confusing. Could be better to use standard `if` condition here?

packages/js/src/Modules/Verto/util/helpers.ts
```diff
@@ -130,3 +131,17 @@ export const isValidOptions = ({
...
+
+  const hasStateParam: GatewayStateType | '' =
+    msg && msg.params && msg.params.state ? msg.params.state : '';
```
arghav, line 142: Same as above, could be more readable with `if`

packages/js/src/Modules/Verto/util/tests/helpers.test.ts
```diff
@@ -0,0 +1,63 @@
+import { getGatewayState } from '../helpers';
+import { IRequestRPC, IResponseRPC } from '../interfaces';
+
+describe('getGatewayState', () => {
+  it('it should return empty string if server did not send the gateway state in the request', () => {
```
arghav, line 5:
```suggestion
  it('request: it should return empty string if server did not send the gateway state', () => {
```

packages/js/src/Modules/Verto/util/tests/helpers.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    const msg: IRequestRPC = {
+      id: '',
+      method: '',
+      params: '',
+    };
+
+    const state = getGatewayState(msg);
+    expect(state).toBeFalsy();
+  });
+  it('it should return empty string if server did not send the gateway state in the response', () => {
```
arghav, line 15:
```suggestion
  it('response: it should return empty string if server did not send the gateway state', () => {
```

arghav: prefixing `request` or `response` makes it a lot more readable since they are all inside the same block. You can also move the tests inside a `describe('request')` and `describe('response')` block.

arghav (approved):
Left a few suggestions that can improve readability.

packages/js/src/Modules/Verto/util/helpers.ts
DeividVeloso: Done!

```
export const getGatewayState = (msg: IMessageRPC): GatewayStateType | '' => {
  let stateResult: GatewayStateType | '' = '';
  let stateParam: GatewayStateType | '' = '';

  if (msg?.result?.params?.state) {
    stateResult = msg?.result?.params?.state;
  }

  if (msg?.params?.state) {
    stateParam = msg?.params?.state;
  }

  const gateWayState = stateResult || stateParam;

  return gateWayState;
};

```

PR #246, opened 2022-01-04
fix(WEBRTC-757): change the timeout value of 500 to use a random number

DeividVeloso:
- Change the timeout value of 500 to use a random number between 3000 and 5000 milliseconds to avoid many users trying to connect to the server at the same time.

## 📝 To Do

- [x] All linters pass
- [ ] All tests pass

## ✋ Manual testing

1. Navigate into `packages/js` and run  `npm run build `
2. Open another project with a web dialer and run `npm link`
3. Try to make many registrations and see if the time change

### Desktop

- [ ] Edge (latest)
- [x] Chrome
- [ ] Firefox
- [ ] Safari

## 📸 Screenshots

| Description | Screenshot |
| ----------- | ---------- |
| Desktop     |image|

### DIFF
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -70,7 +70,7 @@ export default abstract class BaseSession {
   }
 
   get reconnectDelay() {
-    return randomInt(6, 2) * 1000;
+    return randomInt(2, 6) * 1000;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -70,7 +70,7 @@ export default abstract class BaseSession {
   }
 
   get reconnectDelay() {
-    return randomInt(6, 2) * 1000;
+    return randomInt(2, 6) * 1000;
```
arghav, line 73: This could lead to the same issues, won't it? Randomising doesn't guarantee that multiple clients won't connect at the same time.

arghav:
Is the real issue here with multiple clients connecting at the same time? Shouldn't out servers be able to handle that?

packages/js/src/Modules/Verto/BaseSession.ts
DeividVeloso: @StefanYohansson could you reply here, please?

StefanYohansson: this random will decrease the pace. Of course, some can receive the same number, but the problem is not 5~10 connecting at the same time but 120~420 (current values of browsers reconnections).

StefanYohansson (approved):

PR #558, opened 2026-03-19
feat: single-interface ICE candidate filter for multi-NIC DTLS mismatch

lucasassisrosa:
## Problem

On multi-NIC clients (e.g. WiFi + Ethernet), the browser gathers ICE candidates from **both** network interfaces. With trickle ICE, the B2BUA receives and selects a candidate from interface A (first srflx), while the client nominates a candidate from interface B (relay with lower network-cost). This causes a **DTLS path mismatch** — the B2BUA sends DTLS ClientHello to one address, the client expects DTLS on a different address — and the handshake never completes. Result: **zero audio**.

This is intermittent because it only triggers when the client has multiple active network interfaces behind NAT.

## Root Cause

```
Server → sends DTLS to 50.236.171.26:16464 (srflx, interface 1)
Client → waits for DTLS on 64.16.248.199:51648 (relay, interface 2)
DTLS handshake never completes → no SRTP keys → zero audio
```

The B2BUA picks the first srflx candidate passing ACL and sends DTLS there, ignoring the controlling agent's nomination of a different candidate pair. The correct long-term fix is in the B2BUA (respect nomination per RFC 8445), but the SDK should also defend against it.

## Solution

New `CandidateFilter` class that locks trickle ICE candidates to a **single network interface**:

- Every ICE candidate has a `raddr` (related address) = the private IP of the source NIC
- The filter locks to the **first `raddr`** it sees from an srflx/relay candidate
- All subsequent candidates from a different `raddr` are **dropped inline**
- Host candidates always pass through (B2BUA drops them via ACL anyway)
- Single-NIC clients are completely unaffected

### Key properties
- **Zero buffering delay** — candidates processed inline as they arrive
- **Zero customer configuration** — `raddr` identifies the interface automatically
- **All browsers** — `raddr` is standard per RFC 8445, works on Chrome/Firefox/Safari
- **Opt-in flag** — `singleInterfaceIce: true` (default: `false`)

## Usage

```js
new TelnyxRTC({
  trickleIce: true,
  singleInterfaceIce: true,
})
```

## Changes

| File | Description |
|---|---|
| `CandidateFilter.ts` | New class — inline raddr-based interface locking |
| `CandidateFilter.test.ts` | 11 unit tests including exact dual-NIC scenario reproduction |
| `interfaces.ts` (webrtc) | Added `singleInterfaceIce` call option |
| `interfaces.ts` (util) | Added `singleInterfaceIce` session option |
| `BaseCall.ts` | Wired filter into trickle ICE peer events + reset on state reset |

## Testing

- 11 new unit tests covering: disabled passthrough, host/TCP handling, interface locking, raddr=0.0.0.0 fallback, reset behavior, exact dual-NIC scenario reproduction, single-NIC no-op
- All 218 existing tests pass
- TypeScript compiles clean

### DIFF
--- a/packages/js/CHANGELOG.md
+++ b/packages/js/CHANGELOG.md
@@ -1,3 +1,26 @@
+## [2.26.1-beta.4](https://github.com/team-telnyx/webrtc/compare/webrtc/v2.26.0...webrtc/v2.26.1-beta.4) (2026-04-08)
+
+- feat: store source datacenter identifier from REGED message (#583)
+- chore: release webrtc@2.26.1-beta.3 (#599)
+- chore: anchor package.json at 2.26.1-beta.2 for prerelease workflow
+- fix: fall back to network-id when raddr is anonymized (relay-only)
+- refactor: extract magic string to ANONYMIZED_RADDR constant
+- chore: replace Jira links with PR link in code comments
+- refactor: rename filterCandidatesByInterface to singleInterfaceIce
+- feat: add single-interface ICE candidate filter for multi-NIC DTLS mismatch mitigation
+- fix(ci): push release branch before pinning draft release target (#597)
+- fix(ci): pin draft target to bump SHA, publish from tag (#596)
+- fix(ci): single release-it call for bump + tag + draft (#594)
+- fix(ci): single release-it call for bump + tag + draft (#593)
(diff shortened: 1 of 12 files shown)


lucasassisrosa:
closing while dtls fix on core backend is used to address this problem

PR #383, opened 2024-10-08
feat: change reconnection logic

farhat-ha:
[WEBRTC-2307](https://telnyx.atlassian.net/browse/WEBRTC-2307)

Describe your changes

## 📝 To Do

- [ ] All linters pass
- [ ] All tests pass
- [ ] Change documentation based on my changes

## ✋ Manual testing

1. Provide manual testing instructions

### Desktop

- [ ] Edge (latest)
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

## 📸 Screenshots

| Description | Screenshot |
| ----------- | ---------- |
| Desktop     |            |
| usage.gif   |            |

[WEBRTC-2307]: https://telnyx.atlassian.net/browse/WEBRTC-2307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #559, opened 2026-03-19
feat: ICE restart on connection failure

lucasassisrosa:
## Problem

When ICE/DTLS fails during a trickle ICE call (e.g., NAT rebinding, transient network loss, multi-NIC nomination mismatch), the call has zero audio with no recovery mechanism. The existing ICE restart code was commented out in PR #492 by Artem with a FIXME because the SDP exchange after `restartIce()` was never properly implemented.

### History of the FIXME

1. **Oct 2024 (PR #383)** — Haythem added bare `restartIce()` + `onNetworkClose()`. The `restartIce()` was a no-op: it only sets a flag, doesnt create a new offer or exchange SDP.
2. **Dec 2025 (PRs #478+)** — Multiple attempts to make it work with sleep detection and guards, but `startNegotiation()` created a regular offer instead of an ICE restart offer, so fresh credentials were never used.
3. **Jan 5 2026** — `_createIceRestartOffer()` added with proper `createOffer({ iceRestart: true })`, but it raced with the attach/reconnect flow causing conflicts.
4. **Jan 6 2026** — Reverted to bare `restartIce()` to avoid race conditions with attach.
5. **Feb 2 2026 (PR #492)** — Artem commented it out entirely: *"Restart ICE is not working since we do not handle SDP exchange after ICE restart."*

## Solution

Implements ICE restart on connection failure, following [Jitsi's `IceFailedHandling` pattern](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/connectivity/IceFailedHandling.ts):

1. `connectionState` transitions to `failed`
2. Wait configurable grace period (`iceRestartGraceMs`, default 2000ms) for transient recovery
3. If still failed, create new offer with `{ iceRestart: true }` (generates fresh `ufrag`/`pwd`)
4. Send via `telnyx_rtc.invite` with `trickle: true, iceRestart: true`
5. Server answers, `setRemoteDescription` completes the restart
6. New ICE candidates are gathered and trickled normally

### Why this avoids the old problems

The previous implementation lived in `Peer.ts` and raced with the WebSocket reconnect/attach flow. This implementation:
- Lives in **BaseCall** with direct access to `session.execute`
- Only triggers on **outbound trickle ICE calls** — no conflict with attach flow
- Has a **grace period** — no race with initial connection establishment
- Sends a fresh `telnyx_rtc.invite` with `iceRestart: true` — server knows it's a restart, not a new call

### Limitations

- **Outbound calls only.** ICE restart creates a new offer, requiring the SDK to be the offerer. Inbound calls have the SDK as answerer, and we don't have perfect negotiation (role switch) yet. When perfect negotiation is implemented, this can be extended to inbound calls.

### Configuration

```js
new TelnyxRTC({
  trickleIce: true,
  iceRestartGraceMs: 3000, // optional, default 2000ms
})
```

### Safety guards
- Configurable grace period (default 2000ms) for transient recovery
- Max 3 attempts before giving up
- Cancel on connection recovery (resets attempt counter)
- Skip if signaling state closed
- Skip if inbound call (no perfect negotiation)
- Timer cleanup on call teardown

## Changes

| File | Description |
|---|---|
| `BaseCall.ts` | `_scheduleIceRestart`, `_cancelIceRestart`, `_executeIceRestart`; outbound guard; configurable grace; wired into trickle ICE `connectionstatechange`; cleanup in `_finalize` |
| `interfaces.ts` (webrtc) | Added `iceRestartGraceMs` option |
| `interfaces.ts` (util) | Added `iceRestartGraceMs` option |
| `Call.ice-restart.test.ts` | 15 unit tests |

## Testing

- 15 new tests (scheduling, outbound guard, inbound skip, execution, cancellation, error handling, cleanup)
- All 222 tests pass
- TypeScript compiles clean

### DIFF
--- a/packages/js/src/Modules/Verto/tests/webrtc/Call.ice-restart.test.ts
+++ b/packages/js/src/Modules/Verto/tests/webrtc/Call.ice-restart.test.ts
@@ -1,0 +1,314 @@
+// Suppress async errors from peer init lifecycle (mock RTCPeerConnection gets nulled before init completes)
+process.on('uncaughtException', () => {});
+
+import Call from '../../webrtc/Call';
+import Verto from '../..';
+import { VertoMethod } from '../../webrtc/constants';
+import type { IVertoCallOptions } from '../../webrtc/interfaces';
+
+const originalConsoleDebug = console.debug;
+const originalConsoleLog = console.log;
+const originalConsoleGroup = console.group;
+const originalConsoleTable = console.table;
+const originalConsoleGroupEnd = console.groupEnd;
+
+beforeAll(() => {
+  console.debug = jest.fn();
+  console.log = jest.fn();
+  console.group = jest.fn();
+  console.table = jest.fn();
+  console.groupEnd = jest.fn();
+});
+
+afterAll(() => {
+  console.debug = originalConsoleDebug;
+  console.log = originalConsoleLog;
+  console.group = originalConsoleGroup;
+  console.table = originalConsoleTable;
+  console.groupEnd = originalConsoleGroupEnd;
+});
+
+Object.defineProperty(global, 'performance', {
+  writable: true,
+  value: {
+    mark: jest.fn(),
+    measure: jest.fn().mockReturnValue({ duration: 0 }),
+    clearMarks: jest.fn(),
+    clearMeasures: jest.fn(),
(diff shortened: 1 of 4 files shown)


lucasassisrosa:
closing while we don't have perfect negotiation pattern implemented

PR #605, opened 2026-04-13
feat: log region and dc at info level on connect and new call

lucasassisrosa:
### Description

**What:** Add region and datacenter identifiers to SDK console logs at `info` level.

**Why:** SDK logs should clearly display which regional infrastructure is handling the connection. Currently dc/region are only logged at `debug` level (not visible by default). This makes it easy to confirm routing at a glance.

**Log output after this change:**

```
2026-04-13 14:33:11.309 - Connected to Telnyx — region: apac, dc: cn1
2026-04-13 14:33:21.374 - New Call — region: apac, dc: cn1 { destinationNumber: "8004337300", ... }
```

**Changes:**

| File | Change |
|------|--------|
| `VertoHandler.ts` | Log `Connected to Telnyx — region: <region>, dc: <dc>` at info level on REGED |
| `BaseCall.ts` | Log `New Call — region: <region>, dc: <dc>` at info level (replaces plain "New Call with Options") |

Both values come from the gateway REGED response and are already stored on `session.region` / `session.dc`. This just promotes them from debug to info level.

### DIFF
--- a/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
+++ b/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
@@ -2058,7 +2058,10 @@ export default abstract class BaseCall implements IWebRTCCall {
     } else {
       this.setState(State.New);
     }
-    logger.info('New Call with Options:', this.options);
+    logger.info(
+      `New Call — region: ${this.session.region ?? 'unknown'}, dc: ${this.session.dc ?? 'unknown'}`,
+      this.options
+    );
   }
 
   protected _finalize() {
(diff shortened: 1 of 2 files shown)


ArtemPapazian (approved):
Looks good. Clean change that promotes region/dc to info level for better visibility. Two minor notes for awareness (no blockers):

1. The PUNT log reformatting is unrelated to the feature — not a problem, just noting it.
2. The `Connected to Telnyx` info log now fires unconditionally (even when region/dc are absent), showing `unknown` as fallback. Old debug logs only fired when values existed. This is arguably better for visibility, just a behavioral shift worth noting.

PR #607, opened 2026-04-15
Feat: initial restart ice implementation

ArtemPapazian:
## Summary

Implements **ICE restart** for the WebRTC SDK so an active call can recover from a peer connection failure (e.g. network flip, long pause, relay outage) without tearing down the call and re-inviting.

The SDK now detects when `RTCPeerConnection.connectionState` goes to `failed` or `disconnected`, triggers `restartIce()`, and exchanges the renegotiated SDP with the server via Verto `Modify`.

## Why

Previously the SDK had a stubbed-out ICE restart path (commented-out block in `Peer.handleConnectionStateChange`) because there was no SDP exchange wired up for a mid-call renegotiation. On any peer connection failure the call would just surface a warning and effectively be dead until the user hung up or the socket-level Attach flow kicked in.

## SDP exchange flow

```
┌──────────┐                           ┌───────────┐
│  Client  │                           │  Server   │
└────┬─────┘                           └─────┬─────┘
     │                                       │
     │  connectionState = failed /           │
     │  disconnected                         │
     │──► restartIce()                       │
     │──► createOffer() (new local SDP)      │
     │                                       │
     │──► Modify { action: updateMedia,      │
     │            sdp: <new offer> }  ──────►│
     │                                       │
     │                                       │◄─ answer SDP
     │◄──────────────── Modify response ─────│
     │    setRemoteDescription(answer)       │
     │                                       │
     │  connectionState = connected          │
     │──► finishIceRestart()                 │
     │    (flag cleared, retry re-armed)     │
```

## Key changes

**`Peer.ts`**
- New `isIceRestarting` flag + `finishIceRestart()` helper (clears flag + pending timeout).
- `handleConnectionStateChange` now triggers ICE restart on **both** `failed` and `disconnected`, gated on `navigator.onLine` and `session.connected` so the socket reconnect/Attach path still owns the recovery when the WS is down.
- 15s safety timeout on the Modify exchange — if the server never answers, the flag is cleared so we're not stuck in a zombie "restarting" state.
- On `connected`, `_restartedIceOnConnectionStateFailed` is reset so the next failure can trigger another restart (instead of one-shot per Peer lifetime).
- `_isOffer() || isIceRestarting` in `startNegotiation` / `_createOffer` so an answerer-side peer also produces a new offer during restart.
- `resetIceDone` callback wired through the constructor to reset `BaseCall._iceDone` without re-attaching peer event listeners.

**`BaseCall.ts`**
- New `VertoMethod.Modify` handler — applies `params.sdp` via `_onRemoteSdp` (used when the server is the initiator of the renegotiation).
- `_sendIceRestartModify()` sends the offer over Verto `Modify { action: 'updateMedia' }` and applies the answer from the response.
- `_onIceSdp` / `_onTrickleIceSdp` short-circuit to `_sendIceRestartModify` when `isIceRestarting`, so the renegotiated SDP doesn't go out as a new `Invite`/`Answer`.

**`VertoHandler.ts`**
- `Modify` added to the set of methods dispatched to an existing call.

## Testing

- Full existing test suite passes (`yarn test`: 225/225).
- Clean `tsc --noEmit`.
- Manual testing in webdialer: forced peer failure mid-call → `restartIce` → Modify exchange → media resumes without call teardown.

## Notes

- This is a **draft** — leaving room for more manual testing across browsers / trickle-ICE vs non-trickle paths before marking ready for review.
- Reviewer: @lucasassisrosa

### DIFF
--- a/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
+++ b/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
@@ -1128,11 +1128,25 @@ export default abstract class BaseCall implements IWebRTCCall {
       }
       case VertoMethod.Media: {
         performance.mark('telnyx-rtc-media');
+        // Allow Media through during ICE restart (answer SDP for restarted offer)
+        if (this.peer?.isIceRestarting && params.sdp) {
+          logger.info('ICE restart: received Media with answer SDP');
+          this.peer.isIceRestarting = false;
+          this._onRemoteSdp(params.sdp);
+          break;
+        }
         if (this._state >= State.Early) {
           return;
         }
         this.gotEarly = true;
         this._onRemoteSdp(params.sdp);
+        break;
+      }
+      case VertoMethod.Modify: {
+        logger.info('Received Modify message:', params);
+        if (params.sdp) {
+          this._onRemoteSdp(params.sdp);
+        }
         break;
       }
       case VertoMethod.Display: {
@@ -1414,6 +1428,28 @@ export default abstract class BaseCall implements IWebRTCCall {
     return false;
   }
 
+  private _sendIceRestartModify(sdp: string) {
+    const modifyMsg = new Modify({
+      action: 'updateMedia',
+      callID: this.options.id,
+      sdp,
+      dialogParams: this.options,
+    });
+    logger.info('ICE restart: sending Modify with new offer SDP');
+    this._execute(modifyMsg)
+      .then(async (response) => {
+        logger.info('ICE restart Modify response:', response);
+        this.peer.isIceRestarting = false;
+        if (response?.sdp) {
+          await this._onRemoteSdp(response.sdp);
+        }
+      })
+      .catch((error) => {
+        logger.error('ICE restart Modify failed:', error);
+        this.peer.isIceRestarting = false;
+      });
+  }
+
   private async _onRemoteSdp(remoteSdp: string) {
     const sdp = new RTCSessionDescription({
       sdp: remoteSdp,
@@ -1532,6 +1568,12 @@ export default abstract class BaseCall implements IWebRTCCall {
       'User-Agent': `Web-${SDK_VERSION}`,
     };
 
+    // ICE restart: send Modify with new SDP regardless of original call direction
+    if (this.peer?.isIceRestarting) {
+      this._sendIceRestartModify(sdp);
+      return;
+    }
+
     switch (type) {
       case PeerType.Offer:
         this.setState(State.Requesting);
@@ -1611,6 +1653,12 @@ export default abstract class BaseCall implements IWebRTCCall {
       trickle: true,
       'User-Agent': `Web-${SDK_VERSION}`,
     };
+
+    // ICE restart: send Modify with new SDP regardless of original call direction
+    if (this.peer?.isIceRestarting) {
+      this._sendIceRestartModify(sdp);
+      return;
+    }
 
     switch (type) {
       case PeerType.Offer:
--- a/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
+++ b/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
@@ -1864,17 +1864,13 @@ export default abstract class BaseCall implements IWebRTCCall {
     };
 
     instance.onicegatheringstatechange = (event) => {
-      if (this.options.trickleIce) {
-        logger.debug(
-          'ICE gathering state changed:',
-          instance.iceGatheringState,
-          event
-        );
-      }
+      logger.debug(
+        'ICE gathering state changed:',
+        instance.iceGatheringState,
+        event
+      );
       if (instance.iceGatheringState === 'complete') {
-        if (this.options.trickleIce) {
-          logger.debug('Finished gathering candidates');
-        }
+        logger.debug('Finished gathering candidates');
         performance.mark('ice-gathering-completed');
       }
     };
--- a/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
+++ b/packages/js/src/Modules/Verto/webrtc/BaseCall.ts
@@ -277,8 +277,6 @@ export default abstract class BaseCall implements IWebRTCCall {
       this._onPeerConnectionSignalingStateClosed.bind(this);
     this._onTrickleIceSdp = this._onTrickleIceSdp.bind(this);
     this._registerPeerEvents = this._registerPeerEvents.bind(this);
-    this._registerTrickleIcePeerEvents =
-      this._registerTrickleIcePeerEvents.bind(this);
     this._init();
 
     // Create _rings HTMLAudioElement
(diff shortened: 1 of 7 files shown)


packages/js/src/Modules/Verto/webrtc/BaseCall.ts
```diff
@@ -1128,13 +1121,40 @@ export default abstract class BaseCall implements IWebRTCCall {
...
         this._onRemoteSdp(params.sdp);
         break;
       }
+      case VertoMethod.Modify: {
+        logger.info('Received Modify message', {
+          callID: params?.callID,
+          action: params?.action,
+          hasSdp: !!params?.sdp,
+        });
+        if (params.sdp) {
```
lucasassisrosa: why whould it not have sdp? isn't that an exception scenario? if so we need to throw an error

packages/js/src/Modules/Verto/webrtc/BaseCall.ts
```diff
@@ -1414,6 +1434,68 @@ export default abstract class BaseCall implements IWebRTCCall {
     return false;
   }
 
+  private _sendIceRestartModify(sdp: string) {
+    const modifyMsg = new Modify({
+      sessid: this.session.sessionid,
+      action: 'updateMedia',
```
lucasassisrosa: better use constants for this

packages/js/src/Modules/Verto/webrtc/BaseCall.ts
```diff
@@ -1414,6 +1434,68 @@ export default abstract class BaseCall implements IWebRTCCall {
...
+      dialogParams: this.options,
+    });
+    logger.info('ICE restart: sending Modify with new offer SDP');
+    this._execute(modifyMsg)
+      .then(async (response) => {
+        logger.info('ICE restart Modify response received');
+        this.peer?.finishIceRestart();
+        if (response?.sdp) {
+          await this._onRemoteSdp(response.sdp);
+        }
```
lucasassisrosa, line 1433: please log errors on missing sdp and when it's empty

packages/js/src/Modules/Verto/webrtc/BaseCall.ts
```diff
@@ -1796,44 +1892,16 @@ export default abstract class BaseCall implements IWebRTCCall {
...
-      if (screenShare === false) {
-        attachMediaStream(remoteElement, remoteStream);
+      const useTrickle = this.options.trickleIce && !this.peer?.isIceRestarting;
+      if (useTrickle) {
+        this._onTrickleIce(event);
+      } else {
+        if (this.peer?.iceDone) {
+          return;
+        }
+        this._onIce(event);
```
lucasassisrosa, line 1870: why were the methods here on instance removed?

packages/js/src/Modules/Verto/webrtc/Peer.ts
```diff
@@ -97,6 +104,29 @@ export default class Peer {
     this._session = session;
     this._trickleIceSdpFn = trickleIceSdpFn;
     this._registerPeerEvents = registerPeerEvents;
+    // Track offline events independently so ICE restart and Attach never race.
+    // _wasOffline is only cleared on peer `connected`, not on `online`.
```
lucasassisrosa: better rename it to make this clearer. something like wasConnected

lucasassisrosa (changes requested):
Thanks for pushing this through — the direction makes sense, but I found two correctness issues that I think need to be fixed before merge.

1. **Server-initiated `Modify` offer handling does not preserve the non-trickle SDP contract.**

   In `_handleServerModifyOffer()`, we call `createAnswer()`, `setLocalDescription(answer)`, and immediately send `answer.sdp` in `Modify` (`packages/js/src/Modules/Verto/webrtc/BaseCall.ts`). That immediate send is fine for a trickle-ICE flow, but this path is shared with non-trickle calls too.

   For non-trickle, the SDK normally waits for ICE gathering to complete before sending the SDP. Here, a server-initiated `Modify` on a non-trickle call can send an answer before candidates are gathered. The later ICE events are then handled by the normal call setup paths rather than as part of the `Modify` exchange, which can emit `Answer`/`Attach` or standalone candidate messages for a mid-call media update.

   I’d split this by mode: trickle can send the SDP immediately and continue trickling if the backend supports that path, while non-trickle should wait for the complete local SDP and send the answer via `Modify` only once.

2. **`isIceRestarting` can be cleared before the restart offer is actually sent.**

   `handleConnectionStateChange()` sets `isIceRestarting = true` on `failed`/`disconnected` and calls `restartIce()`, but the `connected` branch calls `finishIceRestart()` unconditionally. Because `restartIce()` schedules negotiation asynchronously, a transient `disconnected -> connected` recovery can clear the flag before `onnegotiationneeded` / ICE gathering produces the restart SDP. Once that happens, `_onIceSdp()` / `_onTrickleIceSdp()` no longer short-circuit to `_sendIceRestartModify()` and the mid-call restart SDP can fall through as a normal `Invite`/`Answer`/trickle-candidate flow.

   I’d keep the restart state alive until the Modify exchange has completed or timed out, rather than clearing it on every `connected` transition.

Validation I ran locally:
- `tsc -p packages/js/tsconfig.json --noEmit` ✅
- `jest packages/js/src/Modules/Verto/tests/webrtc/Call.trickle-ice.test.ts packages/js/src/Modules/Verto/tests/webrtc/Call.test.ts --runInBand` ✅
- `rollup -c` from `packages/js` ✅
- full root `jest --runInBand` still has the existing `react-client/src/index.test.tsx` ESM/import parser failure; all 225 Verto tests passed before that suite failed.

packages/js/src/Modules/Verto/webrtc/Peer.ts
```diff
@@ -248,46 +280,51 @@ export default class Peer {
...
+      // ICE restart: fire only when the browser never went offline during
+      // this peer's lifetime. If `window.offline` fired, the Attach flow
+      // owns recovery exclusively. _wasOffline is only cleared on `connected`,
+      // not on the `online` event, so there's no race with BrowserSession's
+      // online handler.
+      if (
+        !this._restartedIceOnConnectionStateFailed &&
+        (connectionState === 'failed' || connectionState === 'disconnected') &&
+        !this._wasOffline &&
+        this._session.connected
```
lucasassisrosa: I'm not sure how the connection is going to be true. we're going to trigger ice restart if connection state is failed or disconnected but _wasOffline  is false, which is when we have network conection?

if so better create a boolean flag to represent this as it may be confusing

lucasassisrosa (commented):
a few concerns

packages/js/src/Modules/Verto/webrtc/BaseCall.ts
ArtemPapazian: If you are about `_registerPeerEvents` and `_registerTrickleIcePeerEvents` then they were basically sharing the same code with tiny specific. So their merge I treat as part of removing duplication. 

Here is the comment -  https://github.com/team-telnyx/webrtc/pull/607/changes#diff-873bf8045359d4688ddcbb64a03d1f4a36886b70c102d34cef104e27ab31c95dR1949-R1959

lucasassisrosa (approved):

PR #614, opened 2026-04-27
fix: prevent double answer when peer connection already exists

lucasassisrosa:
## Problem

The SDK allowed `call.answer()` to be called multiple times on the same call object, creating duplicate PeerConnections. This caused:
- Duplicate `telnyx_rtc.answer` events sent to the server
- Broken media (two negotiations for the same call)
- CallReportCollector tracking the wrong PeerConnection

## Root Cause

Application code (or accidental double event handlers) could invoke `answer()` twice. The SDK had no guard against this.

## Solution

Add an early-return guard in `BaseCall.answer()` that checks if `this.peer.instance` exists and its `signalingState !== 'closed'`. If a peer is already active or connecting, the second `answer()` is ignored and a warning is emitted.

### Changes

- **`BaseCall.ts`**: Guard at the top of `answer()` — returns early with a warning if peer already exists
- **`errorCodes.ts`**: New warning code `ANSWER_WHILE_PEER_ACTIVE: 33006`
- **`warnings.ts`**: Warning metadata with causes and solutions
- **`Call.test.ts`**: Two new tests — one for blocked second answer, one for allowed answer after peer is closed

### Warning Details

| Field | Value |
|---|---|
| Code | 33006 |
| Name | ANSWER_WHILE_PEER_ACTIVE |
| Message | Call answer ignored because a peer connection is already active |

[WEBRTC-3448]: https://telnyx.atlassian.net/browse/WEBRTC-3448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lucasassisrosa:
**Additional context from Bilal Imran on ENGDESK-51431:**

The b2bua-rtc didn't reject the second `telnyx_rtc.answer` either, which completely reset the already established media and caused the OWA. So this is a defense-in-depth fix on the SDK side — ideally the B2BUA should also reject duplicate answers after media is established.

For future reference, the correct ways to update media after a call is answered are:
- `telnyx_rtc.modify` — for mid-call media changes
- `telnyx_rtc.attach` — for reconnection after disconnect (supports trickleICE)

A second `telnyx_rtc.answer` should never be sent once media is active. This guard prevents the SDK from doing so.

cc @ahmed

ArtemPapazian (approved):

PR #622, opened 2026-05-12
fix: prevent duplicate inbound answers

lucasassisrosa:
## Summary
- add a per-credential inbound answer guard so duplicate TelnyxRTC instances in the same JS runtime cannot both answer duplicate inbound call legs
- keep the existing same-call peer guard, and also guard the `_creatingPeer` window before a peer connection is fully active
- emit a new `DUPLICATE_INBOUND_ANSWER` SDK warning when a second inbound answer is ignored
- release the guard when the owning call is destroyed, so future inbound calls can be answered normally

## Context
VSUP-54 showed duplicate WebSocket registrations for the same SDK credential. B2BUA forked the INVITE to both registered contacts, the app answered both SDK call objects, and the losing leg hit SIP 486 USER_BUSY / LOSE_RACE. This keeps the JS SDK from sending the second answer from the same page/runtime.

Checked existing open PRs before starting; none were targeting duplicate answer prevention.

## Validation
- `yarn test packages/js/src/Modules/Verto/tests/webrtc/Call.test.ts --runInBand`
- `yarn workspace @telnyx/webrtc exec eslint src/Modules/Verto/webrtc/BaseCall.ts src/Modules/Verto/util/constants/errorCodes.ts src/Modules/Verto/util/constants/warnings.ts src/Modules/Verto/tests/webrtc/Call.test.ts`
- `yarn workspace @telnyx/webrtc build`
- `git diff --check`


packages/js/src/Modules/Verto/webrtc/BaseCall.ts
```diff
@@ -2059,6 +2067,105 @@ export default abstract class BaseCall implements IWebRTCCall {
...
+      return false;
+    }
+
+    if (
+      ['failed', 'closed', 'disconnected'].includes(
+        peerConnection.iceConnectionState
+      )
+    ) {
+      return false;
+    }
```
ArtemPapazian: Here `failed` and `disconnected` are not fatal. They could be recovered so incoming answer will cause conflict if we will recover existed code. 

I think leaving `closed` would be better. That give us guarantee that existed call is dead.

ArtemPapazian (approved):

packages/js/src/Modules/Verto/webrtc/BaseCall.ts
lucasassisrosa: Addressed in 914f2ab: `failed` and `disconnected` are now treated as recoverable/blocking states; only `closed` peer/signaling states release the duplicate inbound answer guard. Updated the focused test coverage accordingly.

PR #623, opened 2026-05-12
fix: use originator cancel for outbound pre-answer hangup

lucasassisrosa:
## Summary

- use `ORIGINATOR_CANCEL/487` for outbound calls canceled before answer instead of overloading `USER_BUSY/17`
- keep inbound pre-answer rejection and calls without outbound direction as `USER_BUSY/17`

## Context

This is the cause-code half of the original PR #623 split. It targets outbound pre-answer local cancels like:

```text
SEND telnyx_rtc.invite
state requesting -> hangup
SEND telnyx_rtc.bye cause=USER_BUSY causeCode=17
```

For outbound caller-side cancellation before answer, the SDK should send `ORIGINATOR_CANCEL/487`. Inbound pre-answer rejection remains `USER_BUSY/17`.

The separate late invite/answer ACK state-race fix has been moved to its own PR.

## Validation

- `yarn jest packages/js/src/Modules/Verto/tests/webrtc/Call.test.ts --runInBand`

### DIFF
--- a/packages/js/src/Modules/Verto/tests/webrtc/Call.test.ts
+++ b/packages/js/src/Modules/Verto/tests/webrtc/Call.test.ts
@@ -12,7 +12,7 @@ Object.defineProperty(global, 'performance', {
 });
 
 import { isQueued, register, deRegister } from '../../services/Handler';
-import { State } from '../../webrtc/constants';
+import { Direction, State } from '../../webrtc/constants';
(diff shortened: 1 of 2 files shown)


ArtemPapazian (approved):

lucasassisrosa:
this will go out in separate release

PR #625, opened 2026-05-12
fix: log local audio source diagnostics

lucasassisrosa:
## Summary

- Add local audio sender track metadata to call report outbound audio intervals:
  - `id`, `label`, `enabled`, `muted`, `readyState`, `contentHint`
  - selected input settings including `deviceId`, `groupId`, `channelCount`, `sampleRate`, `echoCancellation`, `noiseSuppression`, `autoGainControl`
- Add outbound `media-source` stats to call reports:
  - `trackIdentifier`, `audioLevel`, `totalAudioEnergy`, `totalSamplesDuration`, echo return loss fields when available
- Log a `CallReportCollector: local audio track snapshot` debug entry when local audio track state changes, without repeating it every stats interval.

## Why

Muted/silent outbound investigations currently show RTP bytes/packets increasing, but not whether the local browser track was muted, ended, using the wrong input device, or producing zero-energy source audio.

From two call reports:

- CSV window: 521 client invites from `2026-05-12T12:53:15Z` to `2026-05-12T14:05:30Z`, all `Web-2.25.25`.
- Selected mics varied across the window: Sanas Virtual Audio and many Jabra device instances.
- Silent sample `db85878a-34f7-41ee-9d82-a135dc5affac` used `Microphone (10- Jabra EVOLVE 20) (0b0e:0301)` and had outbound RTP increasing while `audioLevelAvg` stayed `0`.
- Comparison sample `4e570cce-d252-4fcf-a7e8-2de3db0e132a` used `Microphone (Sanas Virtual Audio)` and had non-zero outbound audio initially.

This PR adds the missing endpoint/source evidence needed to distinguish RTP transport issues from wrong/muted/silent local input.

## Validation

- `yarn workspace @telnyx/webrtc build`
- `yarn workspace @telnyx/webrtc test src/Modules/Verto/tests/webrtc/Call.test.ts src/Modules/Verto/tests/webrtc/Call.trickle-ice.test.ts --runInBand`

### DIFF
--- a/packages/js/src/Modules/Verto/tests/webrtc/CallReportCollector.test.ts
+++ b/packages/js/src/Modules/Verto/tests/webrtc/CallReportCollector.test.ts
@@ -1,0 +1,118 @@
+import logger from '../../util/logger';
+import {
+  CallReportCollector,
+  ILocalAudioSourceStats,
+  ILocalAudioTrackSnapshot,
+} from '../../webrtc/CallReportCollector';
+
+type TestableCallReportCollector = {
+  peerConnection: {
+    getSenders: () => Array<{
+      track?: Partial<MediaStreamTrack> & {
+        kind?: string;
+        contentHint?: string;
+        getSettings?: () => Partial<MediaTrackSettings>;
+      };
+    }>;
+  } | null;
+  cleanup: () => void;
+  _withoutUndefined: <T extends Record<string, unknown>>(obj: T) => T;
+  _getLocalAudioTrackSnapshot: () => ILocalAudioTrackSnapshot | undefined;
+  _getOutboundAudioSourceStats: (
+    stats: RTCStatsReport,
+    outboundAudio: RTCOutboundRtpStreamStats & { mediaSourceId?: string }
+  ) => ILocalAudioSourceStats | undefined;
+  _logLocalAudioTrackSnapshot: (
+    localAudioTrack?: ILocalAudioTrackSnapshot,
+    localAudioSource?: ILocalAudioSourceStats
+  ) => void;
+};
+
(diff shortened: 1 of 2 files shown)


ArtemPapazian (approved):
Approving — the change is well-structured, solves a real debugging gap, and the refactor is clean.

**Non-blocking concerns:**

1. **`_withoutUndefined` mutates the input object** — The method deletes keys from the passed object and returns it. It works because callers create fresh objects inline, but the signature `<T extends Record<string, unknown>>(obj: T): T` implies it might return a new object. Consider making it pure: `Object.fromEntries(Object.entries(obj).filter(([_, v]) => v !== undefined))`, or document the mutation explicitly.

2. **JSON.stringify for dedup is order-dependent** — If track property enumeration order varies across browsers, the same logical snapshot could produce different JSON and log again. V8 is stable here, but worth noting for non-Chrome environments.

3. **`_lastLocalAudioTrackSnapshotJson` is never reset** — If the track is replaced mid-call (device switch) and the new track happens to have the same snapshot JSON, the log would be suppressed. Consider resetting it in `cleanup()` for completeness.

4. **No unit tests for new methods** — None of `_getLocalAudioTrackSnapshot`, `_getOutboundAudioSourceStats`, `_logLocalAudioTrackSnapshot`, `_withoutUndefined` have dedicated test coverage. Would like to see tests in a follow-up.

5. **Empty `ILocalAudioTrackSnapshot` edge case** — If all fields happen to be undefined, `_withoutUndefined` would produce `{}` as `localTrack`, and the `if (!localAudioTrack) return` check in `_logLocalAudioTrackSnapshot` would not catch it since `{}` is truthy. In practice track always has id/label, but defensively the log could emit noise.

PR #630, opened 2026-05-14
VSDK-194: Idempotent call cleanup and dead call guard

ArtemPapazian:
## Summary

Implements [VSDK-194](https://linear.app/telnyx/issue/VSDK-194/clean-up-dead-calls-from-the-sdk): Clean up dead calls from the SDK.

### Problem
Two scenarios where calls were not properly cleaned up:

1. **Orphaned calls after reconnection** — When the WebSocket reconnects but the server returns an empty `reattached_sessions` list, the SDK still has active calls that the server no longer knows about. Any call-control operation (hangup, hold, etc.) on these calls will fail with `CALL_DOES_NOT_EXIST`. The calls are unrecoverable and must be terminated locally.

2. **BYE execution hang** — If the BYE message hangs (e.g. socket stalled, server unresponsive), the call would never be cleaned up. A timeout guard now ensures the call is always destroyed after 5 seconds regardless.

### Changes

- **`SESSION_NOT_REATTACHED` error (48501)** — Promoted from warning to fatal error. When the server does not reattach the active session, the call is no longer recoverable. Emitted via `SwEvent.Error` with `createTelnyxError()`.
- **Orphan cleanup via hangup** — Calls that failed to reattach are cleaned up using `call.hangup({}, false)`, which runs the full local cleanup path (state transitions, peer close, stream stop, session removal) without sending BYE to a server that no longer has the session.
- **BYE timeout guard** — If BYE execution hangs for more than 5 seconds, the call proceeds to Destroy so it is always cleaned up. `BYE_TIMEOUT_MS` is a module-level constant.
- **No defensive guards** — Removed `_finalized`, `isFinalized`, and the terminal-state guard from `hangup()`. The underlying operations (`peer.close()`, `stream.stop()`, `delete session.calls[id]`) are all safe to call multiple times. The state machine handles transitions; no extra guards needed.

### Key design decisions
- `SESSION_NOT_REATTACHED` is an error, not a warning — the call is gone, not degraded
- `hangup({}, false)` for orphan cleanup instead of `setState(Destroy)` — follows the same cleanup flow as a normal hangup while skipping BYE for a session that no longer exists
- No ticket keys in runtime logs — public SDK logs should not expose internal ticket numbers
- No `_finalized`/`isFinalized` — session map (`session.calls`) is the source of truth; `hangup()` is safe to call at any time

### Test plan
- [x] Unit tests pass (26/26 — DeadCallCleanup + VertoHandler)
- [ ] Manual test: verify orphaned calls are cleaned up after failed reattach
- [ ] Manual test: verify BYE timeout cleans up call when server is unresponsive
- [ ] Manual test: verify `telnyx.error` event is emitted with `SESSION_NOT_REATTACHED` code

### DIFF
--- a/packages/js/src/Modules/Verto/tests/DeadCallCleanup.test.ts
+++ b/packages/js/src/Modules/Verto/tests/DeadCallCleanup.test.ts
@@ -335,6 +335,8 @@ describe('BaseCall - Dead Call Cleanup (VSDK-194)', () => {
     });
 
     it('should still finalize call when BYE throws an error', async () => {
+      const clearTimeoutSpy = jest.spyOn(global, 'clearTimeout');
+
       call = new Call(session, {
         destinationNumber: '1234',
         callerNumber: '5678',
@@ -359,6 +361,10 @@ describe('BaseCall - Dead Call Cleanup (VSDK-194)', () => {
       // Call should still be cleaned up even though BYE failed
       expect(call.isFinalized).toBe(true);
       expect(session.calls[call.id]).toBeUndefined();
+      // Timeout should be cleared even on rejection path (finally block)
+      expect(clearTimeoutSpy).toHaveBeenCalled();
+
+      clearTimeoutSpy.mockRestore();
     });
 
     it('should clear the BYE timeout when BYE resolves before timeout', async () => {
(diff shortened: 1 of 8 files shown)


ArtemPapazian:
This PR is redundant and split into two other PRs

PR #650, opened 2026-05-21
Fix: signaling health recovery for half-dead WebSocket

ArtemPapazian:
## Problem

When a network interface is removed mid-call while another interface remains, the browser can keep reporting the WebSocket as `OPEN` even though the TCP path bound to the removed interface is effectively half-dead. In that state no inbound signaling frames arrive, ICE restart `Modify` requests can hang, and the call can silently lose media.

The original keepalive/recovery path was too soft and too distributed:
- missed ping/pong activity only logged warnings and did not reliably force reconnect;
- liveness tracking missed JSON-RPC responses consumed by request handlers;
- peer/media failures could trigger signaling probes and ICE restarts from separate paths;
- request timeouts could affect non-critical fire-and-forget requests;
- stale timers/handlers from an old socket could leak into a reconnected socket generation.

## Solution

This PR makes `SignalingHealthMonitor` the single recovery decision authority and keeps socket liveness monitoring separate from media recovery.

### 1. Track all inbound WebSocket activity

- `Connection` updates `lastInboundAt` on every inbound `ws.onmessage`, before JSON parsing/dispatch.
- Emits internal `SocketActivity` for every inbound frame, including JSON-RPC responses consumed by `trigger(msg.id, msg)`.
- Treats even non-JSON/string frames as transport liveness evidence.

### 2. Centralize signaling vs media recovery decisions

`SignalingHealthMonitor` now owns the decision matrix:

- signaling unhealthy/disconnected → socket reconnect path, no ICE restart;
- signaling healthy + peer/media unhealthy → ICE restart only;
- signaling unknown/stale → send one probe and defer ICE restart until inbound activity proves the socket is alive;
- in-flight probe means health is unknown, not healthy or unhealthy.

`Peer` and no-RTP handling now report facts to the monitor instead of directly combining probe/restart behavior themselves.

### 3. Monitor socket liveness for the session lifetime

- The monitor starts on socket open, not only while a call is active.
- Periodic socket liveness checks can send a health probe even when there is no active call.
- Media/peer recovery remains gated by `hasActiveCall()` inside the monitor, so idle sessions do not trigger ICE/media actions.

### 4. Scope request timeouts to critical signaling methods

`BaseSession.execute()` applies `Connection.DEFAULT_REQUEST_TIMEOUT_MS` only for critical methods:

- `telnyx_rtc.modify`
- `telnyx_rtc.bye`
- `telnyx_rtc.ping`

Non-critical/fire-and-forget messages keep the legacy no-timeout behavior and cannot trigger signaling recovery or unhandled timeout rejections.

### 5. Clean up stale request state safely

- `Connection` tracks pending request IDs, timeout handles, and rejecters.
- Timed requests are cleaned up on response, timeout, close, and reconnect.
- Stale-generation timeouts settle with `StaleRequestError` without triggering recovery for a fresh socket.
- Legacy no-timeout requests are deregistered on close/reconnect without forced rejection, avoiding unhandled promise rejections.

### 6. Remove unused/new public surface

Removed unused additions that were no longer needed after centralizing recovery:

- `SwEvent.SignalingHealthProbeRequest`
- `SIGNALING_HEALTH_FAILURE` code/export/registry entry
- `BaseSession.triggerSignalingHealthProbe()`
- `BaseSession.isSignalingHealthy()`
- old direct probe trigger path from `Peer`

## Warning codes

This PR keeps recovery reporting as warnings:

- `SIGNALING_HEALTH_PROBE_TIMEOUT` (`36001`)
- `SIGNALING_REQUEST_TIMEOUT` (`36002`)
- `SIGNALING_RECOVERY_REQUIRED` (`36003`)
- `MEDIA_RECOVERY_REQUIRED` (`36004`)

## Testing

Verified locally:

```bash
yarn workspace @telnyx/webrtc test \
  packages/js/src/Modules/Verto/tests/signaling-health/SignalingHealth.test.ts \
  packages/js/src/Modules/Verto/tests/webrtc/Peer.test.ts \
  packages/js/src/Modules/Verto/tests/webrtc/CallReportCollector.test.ts \
  --runInBand

yarn workspace @telnyx/webrtc build

yarn workspace @telnyx/webrtc eslint \
  src/Modules/Verto/BaseSession.ts \
  src/Modules/Verto/services/Connection.ts \
  src/Modules/Verto/services/SignalingHealthMonitor.ts \
  src/Modules/Verto/tests/signaling-health/SignalingHealth.test.ts \
  src/Modules/Verto/tests/webrtc/Peer.test.ts \
  src/Modules/Verto/webrtc/BaseCall.ts \
  src/Modules/Verto/webrtc/Peer.ts \
  src/Modules/Verto/util/constants/errorCodes.ts \
  src/Modules/Verto/util/constants/errors.ts \
  src/Modules/Verto/util/constants/index.ts \
  src/Modules/Verto/util/interfaces/SignalingHealth.ts
```

Results:
- 64 targeted tests passed.
- WebRTC package build passed.
- ESLint passed on changed files.

## Acceptance criteria

- ✅ Any inbound WS frame updates signaling liveness.
- ✅ Half-dead WS can be detected by inactivity probe and request timeout paths.
- ✅ Peer/no-RTP failures no longer mix socket reconnect and ICE restart decisions.
- ✅ Stale/unknown signaling sends one probe and defers ICE restart until the socket proves healthy.
- ✅ Critical signaling requests cannot hang indefinitely.
- ✅ Non-critical requests keep legacy no-timeout/fire-and-forget behavior.
- ✅ Stale request timers/handlers are cleaned up across close/reconnect without affecting a fresh socket.

## Backward compatibility

- `Connection.send(bladeObj, timeoutMs?)` remains backward compatible; timeout is optional.
- `SocketActivity` remains internal.
- Removed unused new surface before merge rather than exposing it.
- Idle sessions get socket liveness monitoring, but media recovery actions remain active-call scoped.


packages/js/src/Modules/Verto/services/SignalingHealthMonitor.ts
```diff
@@ -0,0 +1,320 @@
...
+        'Signaling health: probe already in flight, skipping triggered probe'
+      );
+      return;
+    }
+
+    const now = Date.now();
+    const silenceMs = now - this._lastInboundAt;
+
+    // If we've had recent activity (< 3s), signaling is likely fine
+    if (silenceMs < 3000) {
```
lucasassisrosa: please move to const this

packages/js/src/Modules/Verto/services/SignalingHealthMonitor.ts
```diff
@@ -0,0 +1,320 @@
...
+  private _probeInFlight: boolean = false;
+  /** The periodic check interval. */
+  private _intervalId: ReturnType<typeof setInterval> | null = null;
+
+  /**
+   * Verto method names that are critical for call control.
+   * Timeouts on these methods indicate the signaling path may be broken
+   * and warrant force-reconnecting the socket.
+   */
+  private static readonly CRITICAL_METHODS = new Set([
```
lucasassisrosa: ping should be here too

lucasassisrosa (commented):
Comment-only review (no approve/reject).

I checked the final diff and ran the local gates:
- `yarn workspace @telnyx/webrtc test --runInBand` ✅ 321 tests
- `yarn workspace @telnyx/webrtc eslint ...changed files...` ✅
- `../../node_modules/.bin/tsc --noEmit --pretty false` ✅
- `yarn workspace @telnyx/webrtc build` ✅

One lifecycle issue worth addressing/confirming:

`Connection.send()` now deregisters the handler and silently returns when the timeout belongs to an old `socketGeneration`. That prevents the stale timeout from force-closing the new healthy socket, which is good, but it also leaves the original Promise permanently pending because the handler was already removed and the timer path neither resolves nor rejects.

A concrete path: ICE restart sends `telnyx_rtc.modify` on the old socket, the health monitor reconnects successfully before the 10s request timeout, then the old timeout takes the stale-generation branch. `BaseCall._sendIceRestartModify().catch(...)` never runs, and any caller awaiting `execute()` also hangs. The peer’s 15s safety timer eventually clears `isIceRestarting`, but the request lifecycle itself remains unresolved.

Suggestion: still settle the original Promise with a stale/cancelled/disconnected error, and have `BaseSession.execute()` skip signaling recovery for that stale-generation case. That keeps the “don’t kill the new socket” behavior without introducing permanently pending request Promises.

packages/js/src/Modules/Verto/webrtc/Peer.ts
```diff
@@ -133,6 +133,57 @@ export default class Peer {
...
+  public restartIce(): boolean {
+    if (!this.instance) {
+      logger.warn('ICE restart: no RTCPeerConnection instance');
+      return false;
+    }
+    if (this.isIceRestarting) {
+      logger.debug('ICE restart: already in progress, skipping');
+      return false;
+    }
+    // Do not restart ICE if the browser went offline — the Attach flow
```
lucasassisrosa: I'm reluctant on this offline detection to trigger restartIce or not -- this browser offline can be false positive

lucasassisrosa:
## Review

Consolidates WebSocket liveness, request-timeout, and ICE/media-failure handling under a single `SignalingHealthMonitor`. The core invariant is sound:

- **Signaling unhealthy → socket reconnect only** (never ICE restart on a dead path)
- **Signaling healthy + media unhealthy → ICE restart only** (never reconnect)
- **Unknown → probe once, defer media recovery until inbound activity proves the socket is alive**

Plus `Connection.socketGeneration` to prevent stale request timeouts from affecting a fresh socket. The architectural direction is the right one — recovery decisions were previously scattered and could race.

### Strengths

- **Clear separation of concerns.** `Peer` now reports facts (`reportPeerFailure`/`reportNoRtp`) and the monitor decides — no more peer-initiated probes mixing with reconnect logic.
- **Socket generation guards** (`Connection.ts` ~340-347, ~447-462) elegantly prevent a stale 10s timer from force-closing a freshly reconnected socket — a real race that the comment in `_sendProbe` explicitly documents.
- **Pre-parse activity tracking** (`Connection.ts` ~485-507) is correct: any inbound bytes prove TCP liveness, even non-JSON frames. The comment justifies this well.
- **Critical-vs-non-critical method allowlist** (only `Modify`/`Bye`/`Ping` trigger force-reconnect) protects against fire-and-forget timeouts triggering disruptive recovery.
- **Good test coverage** of the decision matrix, stale-generation handling, and pre-parse activity edge cases.

### Issues

#### 1. `_recoveryGeneration` is dead code — `SignalingHealthMonitor.ts`
The counter is incremented in `_triggerSignalingRecovery` and `_triggerIceRestart` but **never read anywhere**. The doc comment claims it "prevents duplicate recovery from stale events," but no check uses it. Either wire it into a guard (compare against generation at event time) or remove the counter and update the comment.

#### 2. Behavior change in `Peer.restartIce()` is not flagged
The old `_onConnectionStateChange` path guarded ICE restart with `!this._restartedIceOnConnectionStateFailed` — **at most one ICE restart per peer lifetime**. The new public `restartIce()` only checks `isIceRestarting` (with a 30s timeout reset). It still *sets* `_restartedIceOnConnectionStateFailed = true` but no longer reads it, so the monitor can trigger ICE restart repeatedly. This may be intentional (centralized recovery makes the old guard unnecessary), but it's not called out in the PR description. Worth verifying that repeated `restartIce` calls won't churn during a flapping network.

#### 3. Redundant call in `_onSocketOpen` — `BaseSession.ts`
```ts
this._signalingHealthMonitor.onSocketActivity();
this.startSignalingHealthMonitor();
```
`start()` resets `_lastInboundAt = Date.now()` itself, so the prior `onSocketActivity()` is overwritten. Drop the first call.

#### 4. `_pendingMediaRecovery` is a single slot — `SignalingHealthMonitor.ts`
If a peer failure for call-A is pending and a peer failure for call-B arrives while the probe is still in flight, A is silently overwritten. Unlikely with single-call usage but worth either documenting or switching to `Map<callId, …>` if multi-call is supported.

#### 5. Test `'stale request errors settle without invoking signaling recovery'` is hollow — `SignalingHealth.test.ts`
It constructs a `StaleRequestError` and asserts `triggerIceRestart`/`socketDisconnect` weren't called, but never feeds the error through any code path that *could* have called them. The assertion would pass even if the error handler were broken. Either remove this test or route the error through `session.execute().catch(...)` to actually exercise the guard in `BaseSession.execute`.

#### 6. Magic-number duplication in tests — `SignalingHealth.test.ts`
`RECENT_ACTIVITY_THRESHOLD_MS_FOR_TEST = 3_000` shadows the constant inside `SignalingHealthMonitor.ts`. Export it (or a `__test__` symbol) so the test breaks correctly if the production threshold changes.

#### 7. Probe `Ping` promise hangs forever — `SignalingHealthMonitor._sendProbe`
`connection.send(new Ping(...))` is sent without a timeout (by design — the 5s check in `_check()` is the timeout). The handler stays in `_pendingRequestIds` until either response, socket close, or reconnect — fine. But the promise itself never settles in the timeout path. The `.catch()` only catches send failure, not non-resolution. Combined with the no-timeout cleanup path that intentionally doesn't reject (`Connection._cleanupPendingRequests`), this is consistent — but `_sendProbe` should add a comment noting the promise is intentionally orphaned.

### Smaller notes

- `restartIce()` in `Peer.ts` duplicates the connection/offline guards already in `_recoverMediaOrSignaling`. Defensive but worth a comment.
- `BaseCall.ts` ~1166-1167: `this.session.startSignalingHealthMonitor()` is called on every state transition that reaches the `connected` branch. `start()` is idempotent, but the call could be moved to the actual `Active` transition rather than on every reconnect-cycle pass.
- The optional-chaining `session.reportNoRtp?.(...)` / `session.reportPeerFailure?.(...)` in `BaseCall.ts` and `Peer.ts` suggests the methods may not always be present — but they're declared in `ISignalingHealthSession` as required and added to `BaseSession`. Drop the `?.` once the interface is the source of truth.
- The PR removes `bytesSent: 1000` from `createStatsEntry` in `CallReportCollector.test.ts` — silently changes the meaning of unrelated tests using that helper. Worth confirming none rely on the previous value.

### Risks

- **Public-surface increase**: 7 new methods on `BaseSession`, 4 new warning codes, 1 new `SwEvent`. Once exposed they're hard to retract — consider whether `triggerIceRestart`/`reportPeerFailure`/`reportNoRtp` should be marked `@internal` or moved off `BaseSession` since they're really a monitor-internal callback channel.
- **Type erosion**: heavy `this as unknown as {...}` casting in `BaseSession.hasActiveCall`/`triggerIceRestart` is a code smell. The interface `ISignalingHealthSession` exists — why not also a `SessionWithCalls` interface to type the `calls` map properly?
- **Behavioral risk during flapping networks**: with repeated ICE restarts now possible per peer lifetime (issue #2), a network that briefly drops and recovers in a 30s window could churn ICE restart → recover → ICE restart. Suggest adding a soft cooldown in `_triggerIceRestart`.

### Verdict

Architecturally a net win — the decision matrix and socket-generation guard close real bugs. Recommend addressing **#1 (dead `_recoveryGeneration`)**, **#2 (document the lifted ICE-restart cap)**, **#3 (redundant `onSocketActivity` call)**, and **#5 (hollow test)** before merging. The rest can be follow-ups.

packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -150,7 +166,44 @@ export default abstract class BaseSession {
...
+    return sendPromise.catch(async (error) => {
+      // Stale request — socket was replaced before the timeout fired.
+      // Settle the promise but do NOT trigger signaling recovery; the
+      // new socket is healthy and should not be force-closed.
+      if (error instanceof StaleRequestError) {
+        logger.debug(
+          `Stale request settled (id=${error.requestId}, gen=${error.staleGeneration}) — not triggering recovery`
+        );
+        throw error;
+      }
```
lucasassisrosa, line 197: coulld you detail this case? not clear when it's StaleRequestError

packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -918,6 +982,124 @@ export default abstract class BaseSession {
...
+    }
+    const peer = call.peer;
+    if (!peer?.restartIce) {
+      logger.warn(
+        `Signaling health: cannot trigger ICE restart — no peer for call ${callId}`
+      );
+      return;
+    }
+    const started = peer.restartIce();
+    if (!started) {
```
lucasassisrosa: do we have access to reason for this? it'd be good to log the reason

packages/js/src/Modules/Verto/services/Connection.ts
```diff
@@ -148,6 +181,9 @@ export default class Connection {
 
     try {
       this._wsClient = new WebSocketClass(websocketUrl.toString());
+      this.socketGeneration += 1;
```
lucasassisrosa, line 185: it'd be good to log a warning here, on current socket, if socket generation increases abnormally

packages/js/src/Modules/Verto/services/Connection.ts
```diff
@@ -165,17 +201,77 @@ export default class Connection {
...
+        reject,
+        generation: currentGeneration,
       });
+
+      if (timeoutMs && timeoutMs > 0) {
+        timerId = setTimeout(() => {
+          timedOut = true;
+          timerId = null;
+          this._pendingRequestTimers.delete(request.id);
+          this._pendingRequestRejecters.delete(request.id);
```
lucasassisrosa, line 254: better look for a lib to handle these timers

packages/js/src/Modules/Verto/services/SignalingHealthMonitor.ts
```diff
@@ -0,0 +1,505 @@
...
+      );
+      this._sendProbe();
+      return;
+    }
+
+    // Probe is in flight — check if it has timed out
+    const probeElapsedMs = now - this._lastProbeSentAt;
+    if (probeElapsedMs >= PROBE_TIMEOUT_MS) {
+      logger.warn(
+        `Signaling health: probe timed out after ${probeElapsedMs}ms with no inbound activity — declaring signaling unhealthy`
```
lucasassisrosa, line 332: probe should be based on message received from b2bua-rtc via socket, the messag with the same id

lucasassisrosa (commented):
some concerns, mainly about the probe one

packages/js/src/Modules/Verto/webrtc/Peer.ts
ArtemPapazian: Yes, it could be redundant because before triggering restartIce we ensure that socket is alive.

packages/js/src/Modules/Verto/BaseSession.ts
ArtemPapazian: I added note, but we shouldn't see it much or at all:
`Stale request — a request was sent on an old socket generation,
  then reconnect created a replacement socket before the old request
  timeout/cleanup settled. Surface the stale cancellation to the
  caller, but do NOT trigger signaling recovery; otherwise an old
  Modify/Ping timeout could force-close the new healthy socket.`

packages/js/src/Modules/Verto/webrtc/Peer.ts
ArtemPapazian: Addressed in `6c7609a`: removed the sticky browser `offline` event gate from `Peer.restartIce()`, so a transient/false-positive offline event no longer blocks ICE restart after signaling health is confirmed.

packages/js/src/Modules/Verto/BaseSession.ts
ArtemPapazian: Addressed in `6c7609a`: expanded the `StaleRequestError` comment to describe the old-socket-generation timeout path, and added coverage through `BaseSession.execute()` to verify stale errors do not trigger signaling recovery.

packages/js/src/Modules/Verto/BaseSession.ts
ArtemPapazian: Addressed in `10d36f2`: `Peer.restartIce()` now returns a structured result with skip reason, and `BaseSession.triggerIceRestart()` logs that reason when the restart is skipped.

packages/js/src/Modules/Verto/services/Connection.ts
ArtemPapazian: Leaving this as-is intentionally: `socketGeneration` increments on each `connect()`, so a warning without a defined threshold/window would be noisy. The stale-generation behavior is covered by the request-generation tests.

packages/js/src/Modules/Verto/services/Connection.ts
ArtemPapazian: Leaving explicit timers here intentionally: the timer lifecycle is scoped to one request handler and cleanup paths are covered by tests. Adding a timer dependency for this small ownership model would add more surface area than it removes.

packages/js/src/Modules/Verto/services/SignalingHealthMonitor.ts
ArtemPapazian: Addressed in `6c7609a`: a health probe now resolves only through the matching `Ping` request response handled by `Connection.send()`. Generic inbound socket activity only updates passive liveness and no longer releases pending media recovery; tests cover unrelated activity vs. matching probe response.

lucasassisrosa (approved):

packages/js/src/Modules/Verto/services/Connection.ts
lucasassisrosa: just so we know how many sockets open a client has

ArtemPapazian: Added debug logging in `Connection.connect()` with `sessionId`, `voiceSdkId`, `socketGeneration`, and `reconnectCount` so we can see how many reconnects already happened for that client. Commit: 3623cee. Validation: `yarn test src/Modules/Verto/tests/services/Connection.test.ts --runInBand` and `yarn eslint src/Modules/Verto/services/Connection.ts src/Modules/Verto/tests/services/Connection.test.ts`.

PR #664, opened 2026-06-01
fix: default audio codec preference to opus

lucasassisrosa:
## Summary
- Default browser audio codec preferences to Opus first, with remaining browser-supported audio codecs as fallbacks when `preferred_codecs` is omitted.
- Propagate client/session-level `preferred_codecs` into inbound call options.
- Filter answer-side codec preferences to codecs actually present in the remote SDP offer before calling `setCodecPreferences`.
- Emit `telnyx.warning` for codec edge cases:
  - `CODEC_PREFERENCES_UNSUPPORTED` when the browser/transceiver cannot apply preferences.
  - `PREFERRED_CODECS_UNAVAILABLE` when requested/default preferred codecs are partially or fully absent from an inbound remote SDP offer.
- Update preferred-codec docs to clarify the default and the inbound-offer limitation.

## Git history context checked
- `cac97ac` / #361 introduced `preferred_codecs` as caller codec preference and the follow-up commit in the same PR explicitly added answer-side `setCodecPreferences` (`feat: set codec preferences on answer`). So history indicates the feature was intended to work for both outbound offers and inbound answers, but only when `preferred_codecs` was explicitly passed.
- `14c1287`, `e0299f9`, and `57bda0c` only updated preferred-codec docs/examples; they did not add an SDK default or narrow the behavior to outbound-only.
- `4324c69` / #549 reused `getPreferredCodecs` for recvonly audio, so making the helper default to Opus also covers receive-only AI/audio flows.

## Validation
- `yarn workspace @telnyx/webrtc test packages/js/src/Modules/Verto/tests/webrtc/Peer.test.ts --runInBand` ✅ 5 passing
- `yarn workspace @telnyx/webrtc test --runInBand` ✅ 358 passing
- `yarn workspace @telnyx/webrtc build` ✅
- `yarn workspace @telnyx/webrtc lint` ⚠️ still fails on pre-existing repository lint debt; README-specific lint errors were fixed, only existing README table warnings remain.


ArtemPapazian (approved):

lucasassisrosa:
this is depending on more validation to confirm behavior is expected

PR #666, opened 2026-06-02
Fix: reattach session handling on reconnect

ArtemPapazian:
## Summary

Splits the reattach-handling logic out of PR #630 into a standalone draft PR.

**Scope: reattach sessions only** — no BYE timeout and no broader lifecycle cleanup/refactor.

## What this does

When the WebSocket reconnects and the server sends `reattached_sessions` in the `clientReady`/`REGED` message, the SDK now reconciles its active calls against the server's known sessions:

1. **Active call + empty `reattached_sessions`** → terminate the local active call with `SESSION_NOT_REATTACHED` error. Local cleanup only; no BYE sent because b2bua-rtc no longer has the session.

2. **Active call + `reattached_sessions` doesn't contain it** → terminate the SDK active call with the same error. Log + emit warning for the unexpected arrived session(s). Do **not** establish the arrived call automatically.

3. **Active call + `reattached_sessions` contains the matching session** → keep the active call. If there are extra sessions in the array, log + emit warning about the extras, but do not establish them.

4. **No active call + exactly one session** → no action needed; the server will send an `attach` message to restore the session (page reload recovery).

5. **No active call + 2+ sessions** → log + emit `AMBIGUOUS_REATTACHED_SESSIONS` warning; SDK cannot safely choose, so no sessions are established automatically.

## Key changes

- **`SESSION_NOT_REATTACHED` promoted from warning (35001) to error (48501)** — the call is gone, not degraded. Emitted via `telnyx.error` event.
- **New `AMBIGUOUS_REATTACHED_SESSIONS` warning (35002)** — for unexpected/ambiguous arrived sessions that don't match the active SDK call, or when multiple sessions arrive with no active call.
- **Orphan cleanup via `hangup({}, false)`** — follows the same cleanup flow as a normal hangup while skipping BYE for a session that no longer exists server-side.
- **Attach suppression via `_ignoredReattachSessionIds`** — unexpected/ambiguous sessions are tracked in an ignored set; subsequent `Attach` messages for those sessions are ACK'd without answering, preventing auto-establishment.
- **Detailed logging** — includes active SDK call/session ID, `reattached_sessions.length`, arrived call/session IDs, and chosen action (`keep_active`, `terminate_active_missing_reattach`, `ignore_ambiguous_sessions`).

## Test plan

- [x] Unit tests pass (32/32 VertoHandler tests including 12 reattach scenario tests)
- [ ] Manual test: verify orphaned calls are cleaned up after failed reattach
- [ ] Manual test: verify active call is kept when reattached_sessions matches
- [ ] Manual test: verify ambiguous sessions warning is emitted correctly
- [ ] Manual test: verify Attach is suppressed for ignored sessions
- [ ] Manual test: verify BYE is not sent during orphan cleanup

## Files changed

- `VertoHandler.ts` — full reattach logic (5 scenarios) + Attach suppression for ignored sessions
- `errorCodes.ts` — SESSION_NOT_REATTACHED → error 48501, added AMBIGUOUS_REATTACHED_SESSIONS warning 35002
- `errors.ts` — 48501 error definition
- `warnings.ts` — replaced SESSION_NOT_REATTACHED warning with AMBIGUOUS_REATTACHED_SESSIONS
- `VertoHandler.test.ts` — 12 reattach test cases covering all scenarios, BYE skip, Attach suppression, and ignored set reset


lucasassisrosa (approved):

packages/js/src/Modules/Verto/webrtc/VertoHandler.ts
```diff
@@ -72,21 +73,41 @@ class VertoHandler {
...
-      Array.isArray(params?.reattached_sessions) &&
-      params.reattached_sessions.length === 0 &&
-      Object.keys(session.calls).length > 0
-    ) {
-      const warning = createTelnyxWarning(SESSION_NOT_REATTACHED);
-      trigger(
-        SwEvent.Warning,
-        { warning, sessionId: session.sessionid },
-        session.uuid
+    const activeCallsIDs = new Set(Object.keys(session.calls));
```
lucasassisrosa, line 76: before the operations below, I think it's useful to log debug the active calls ids for troubleshooting

ArtemPapazian: Addressed in 29287c6: added a debug log that prints active call IDs before the reattach cleanup check.

PR #667, opened 2026-06-03
Fix: preserve call mic mute state across local media changes

ArtemPapazian:
### Problem

`muteAudio()` only sets `track.enabled = false` on the current stream. There is no durable "I want to be muted" state. When local audio tracks are replaced (device switch, reattach, ICE restart/renegotiation), the new tracks start enabled, losing the mute. `isAudioMuted` reads from `track.enabled` which can diverge from what the user expects.

This is the root cause of VSDK-205: a supervisor calling `muteAudio()` has the mute lost after ~2-4 seconds when ICE renegotiation or reattach replaces/re-enables tracks.

### Solution

Add a durable call-level desired mute state (`_desiredAudioMuted`) initialized from `mutedMicOnStart` that persists across all track replacement scenarios:

- **`muteAudio()`** — sets `_desiredAudioMuted = true` + disables tracks
- **`unmuteAudio()`** — sets `_desiredAudioMuted = false` + enables tracks
- **`isAudioMuted`** — returns `_desiredAudioMuted` (durable state, not track state)
- **`setAudioInDevice()`** — defaults `muted` to `_desiredAudioMuted`; explicit arg updates it
- **`setVideoDevice()`** — applies `_desiredAudioMuted` after stream replacement
- **Peer init** — calls `applyDesiredAudioMuteState` callback after creating local stream
- **VertoHandler Attach** — propagates existing call's `isAudioMuted` as `mutedMicOnStart` for reattached calls

### Files Changed

- `packages/js/src/Modules/Verto/webrtc/BaseCall.ts` — core logic
- `packages/js/src/Modules/Verto/webrtc/Peer.ts` — apply callback after stream setup
- `packages/js/src/Modules/Verto/webrtc/VertoHandler.ts` — reattach mute propagation
- `packages/js/src/Modules/Verto/webrtc/interfaces.ts` — `applyDesiredAudioMuteState` callback type
- `packages/js/src/Modules/Verto/tests/webrtc/AudioMuteState.test.ts` — 23 new tests

### Test Results

All 376 tests pass (including 23 new ones for mute state preservation).


packages/js/src/Modules/Verto/webrtc/BaseCall.ts
```diff
@@ -891,41 +957,98 @@ export default abstract class BaseCall implements IWebRTCCall {
...
+    if (!sender) {
+      // No audio sender — nothing to replace. Keep the current desired state.
+      logger.debug(
+        'Skipping audio input device change: no audio sender found',
+        {
+          callId: this.id,
+          deviceId,
+          audioMuted: this._desiredAudioMuted,
+          audioTrackId: this._getLocalAudioTrackId(),
+        }
```
lucasassisrosa: this should be a warning emitted

packages/js/src/Modules/Verto/webrtc/VertoHandler.ts
```diff
@@ -120,7 +120,8 @@ class VertoHandler {
 
     const _buildCall = (
       recoveredCallId?: string,
-      forceRelayCandidateForRecovery?: boolean
+      forceRelayCandidateForRecovery?: boolean,
+      mutedMicOnStartFromRecovery?: boolean
```
lucasassisrosa: I don't see this option being defined elsewhere

lucasassisrosa: I mean buildCall is called but I don't see these params passed. Maybe we need to change params signature to be an object (required) and define defaults to have bettter control over it

lucasassisrosa (approved):

PR #670, opened 2026-06-03
Fix: handle TIMEOUT gateway state with auto-reconnect

ArtemPapazian:
## What changed

- **Add `TIMEOUT` to `GatewayStateType` enum** — the server can return this state when the SIP registrar is unresponsive
- **Route `TIMEOUT` through the same terminal gateway-failure recovery path** as `FAILED` / `FAIL_WAIT` — emits `GATEWAY_FAILED` (45004) and attempts auto-reconnect
- **Set `skipLastVoiceSdkId = true` on gateway failure** — when the connection fails due to FAILED, FAIL_WAIT, or TIMEOUT, the SDK now requests a different b2bua-rtc instance on reconnect instead of sticky-routing to the same unresponsive target
- **Update `GATEWAY_FAILED` error description** to include TIMEOUT state
- **Add unit tests** for TIMEOUT gateway state handling

## Why

When the SIP registrar times out, the server returns a `TIMEOUT` gateway state. The client was not handling this state, leaving sessions stuck without recovery. This routes TIMEOUT through the same proven recovery path used for FAILED/FAIL_WAIT, and ensures reconnect attempts avoid the same broken b2bua-rtc instance.

## Linear

[VSUP-84](https://linear.app/telnyx/issue/VSUP-84/handle-timeout-and-failed-gateway-states-on-client)

## Test plan

- [x] Unit tests pass (354/354)
- [ ] TIMEOUT triggers `GATEWAY_FAILED` + auto-reconnect
- [ ] TIMEOUT with autoReconnect disabled emits terminal `RECONNECTION_EXHAUSTED`
- [ ] FAILED/TIMEOUT recovery does not reuse the same `voice_sdk_id`
- [ ] No reconnect-attempt reset occurs until confirmed `REGED`

### DIFF
--- a/packages/js/src/Modules/Verto/webrtc/VertoHandler.ts
+++ b/packages/js/src/Modules/Verto/webrtc/VertoHandler.ts
@@ -289,7 +289,12 @@ class VertoHandler {
             // If the user is REGED tell the client that it is ready to make calls
             case GatewayStateType.REGISTER:
             case GatewayStateType.REGED: {
-              if (!this.isDuplicateGatewayState(gateWayState)) {
+              if (
+                !this.isDuplicateGatewayState(gateWayState, [
+                  GatewayStateType.REGED,
+                  GatewayStateType.REGISTER,
+                ])
+              ) {
                 this.session._triggerKeepAliveTimeoutCheck();
                 this.retriedRegister = 0;
 
@@ -370,7 +375,13 @@ class VertoHandler {
             case GatewayStateType.FAILED:
             case GatewayStateType.FAIL_WAIT:
             case GatewayStateType.TIMEOUT: {
-              if (!this.isDuplicateGatewayState(gateWayState)) {
+              if (
+                !this.isDuplicateGatewayState(gateWayState, [
+                  GatewayStateType.FAILED,
+                  GatewayStateType.FAIL_WAIT,
+                  GatewayStateType.TIMEOUT,
+                ])
+              ) {
                 // Emit gateway failure on first occurrence
                 const gatewayError = createTelnyxError(
                   GATEWAY_FAILED,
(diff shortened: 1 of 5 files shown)


packages/js/src/Modules/Verto/webrtc/VertoHandler.ts
```diff
@@ -373,12 +373,15 @@ class VertoHandler {
...
+            case GatewayStateType.FAIL_WAIT:
+            case GatewayStateType.TIMEOUT: {
               if (
                 session.connection.previousGatewayState !==
                   GatewayStateType.FAILED &&
                 session.connection.previousGatewayState !==
-                  GatewayStateType.FAIL_WAIT
+                  GatewayStateType.FAIL_WAIT &&
+                session.connection.previousGatewayState !==
+                  GatewayStateType.TIMEOUT
```
lucasassisrosa: I think on the timeout case, we don't need to check if the last state was timeout too

lucasassisrosa: better revisit why it's checking previous gateway state here

packages/js/src/Modules/Verto/webrtc/VertoHandler.ts
```diff
@@ -394,11 +397,15 @@ class VertoHandler {
                   session.uuid
                 );
 
+                // Avoid sticky reconnect to the same b2bua-rtc target by
+                // requesting a different instance on the next connect().
+                session.options.skipLastVoiceSdkId = true;
```
lucasassisrosa, line 401: please log this since you're updating session.options on the fly we need to track this in call/session reports

lucasassisrosa (approved):

packages/js/src/Modules/Verto/webrtc/VertoHandler.ts
ArtemPapazian: This [PR](https://github.com/team-telnyx/webrtc/pull/235/changes) where the logic was made.  I would leave this without changes. 

But I would improve this checking and log when double state appears.

PR #671, opened 2026-06-09
fix(webrtc): add null guards to connection state event handlers

itxmoizdev:
## Summary
This PR adds guard clauses to `handleConnectionStateChange` and `_handleIceConnectionStateChange` to prevent potential `TypeError` crashes during call teardown.

## Why this is needed
In mobile and React Native environments, WebRTC events can fire asynchronously during or after call teardown. When `this.instance` is set to `null` but an event is still processed, the SDK attempts to access properties on a null object, which causes the host application to crash.

## Changes
- Added null checks at the beginning of `handleConnectionStateChange` and `_handleIceConnectionStateChange`.
- Added debug logging to identify when these events are triggered post-teardown.

## Testing
- Verified that call teardown no longer triggers unhandled exceptions in the event handlers.
- Confirmed that debug logs correctly capture the ignored events during teardown.

### DIFF
--- a/packages/js/src/Modules/Verto/webrtc/Peer.ts
+++ b/packages/js/src/Modules/Verto/webrtc/Peer.ts
@@ -317,12 +317,6 @@ export default class Peer {
   }
 
   private handleConnectionStateChange = async () => {
-    // Guard Clause: Prevent crashes if the peer connection instance is already null during teardown
-    if (!this.instance) {
-      logger.debug('Connection state change ignored: instance is null');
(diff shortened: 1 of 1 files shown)


itxmoizdev:
Hi everyone, I've added these guard clauses to prevent null reference crashes during call teardown in my React Native environment. Looking forward to your feedback!

lucasassisrosa (approved):

ArtemPapazian (approved):

PR #714, opened 2026-06-30
feat(sdk): persist and retry call reports on WebSocket 1001 disconnect

lucasassisrosa:
## Problem

When a WebSocket closes with code 1001 (Going Away) while a call is active, the call report is permanently lost because:

1. **HTTP POST fails** — the call report is POSTed to `/call_report` on voice-sdk-proxy, but when the network is down alongside the socket, the POST also fails.
2. **Report lives only in in-memory state** — `keepConnectionAliveOnSocketClose: false` discards all connection/call state on socket close, including the pending report data.
3. **Auto-reconnect creates a fresh session** — the reconnected session has no knowledge of the orphaned call's pending report. The server assigns a new `call_report_id` at login, but the SDK never gets to populate and submit the orphaned one.

**Investigation evidence** (user 252, peer_ip 186.192.96.25, 2026-06-29 23:44–23:53 UTC):
- Two calls orphaned by `{:remote, 1001, ""}` close — no BYE, no call report submitted
- Call `3f7dad43` (socket 48911): ~87s active, died at 23:51:26
- Call `e1e2bcf0` (socket 48040): ~90s active, died at 23:53:03
- Auto-reconnect (3s later) re-established WebSocket + login only; did NOT restore call state or retry the pending report

## Solution

This PR implements a **sessionStorage-backed persistence + retry mechanism** for call reports that could not be delivered due to socket disconnect, network failure, or connect timeout.

### Changes

**`CallReportStorage.ts`** (new) — sessionStorage-backed pending report queue:
- `enqueuePendingReport()` — stores failed report payloads up to 20 entries / 4 MiB, dropping oldest when exceeded
- `drainPendingReports()` — atomic read + clear (at-most-once delivery semantics)
- Safe try/catch wrappers mirroring `reconnect.ts` pattern — private mode / quota exceeded → silent no-op

**`CallReportCollector.ts`** — extracted `buildFinalPayload(summary)`:
- Returns `ICallReportPayload | null` without sending, enabling callers to persist the payload on POST failure
- `postReport()` now delegates to `buildFinalPayload()` then `_sendPayload()`

**`BaseCall.ts`** — `_flushIntermediateReport()` and `_postCallReport()`:
- Generate synthetic `gen-{uuid}` callReportId when the server hasn't assigned one (socket never connected, REGED not received, or ID lost on reconnect)
- Fall back to `this.session.options.host` when `connection.host` is null (socket closed / never connected)
- On POST failure, persist the payload to sessionStorage via `enqueuePendingReport()` for later retry

**`BaseSession.ts`** — `_onSocketOpen()`:
- Calls `_drainPendingReports()` which reads pending reports from sessionStorage and POSTs each one via fire-and-forget `fetch()` to the `/call_report` HTTP endpoint
- At-most-once delivery: storage is cleared on read, no re-enqueue on failure (prevents unbounded growth in prolonged outage)

### Mechanism summary

```
Socket 1001 close → HTTP POST fails → enqueuePendingReport(sessionStorage)
                                              ↓
Auto-reconnect → _onSocketOpen() → _drainPendingReports() → POST to /call_report
                                              ↓
                                         At-most-once (storage cleared on read)
```

### Scenarios covered

| Scenario | Before | After |
|---|---|---|
| Socket 1001 close mid-call | Report lost | Persisted to sessionStorage, retried on reconnect |
| Socket never connects | Report lost (no callReportId) | Synthetic `gen-` ID generated, report submitted via fallback host |
| Socket connect timeout | Report lost | Same — synthetic ID + fallback host + persistence |
| Prolonged outage (>20 reports) | N/A | Oldest reports dropped (bounded storage) |
| Private mode / storage blocked | N/A | Silent no-op, no crash |

### Tests

- **`CallReportStorage.test.ts`** (new) — 11 tests: enqueue/drain/count/clear, FIFO order, MAX_PENDING_REPORTS limit, malformed data handling, at-most-once semantics
- **`ReconnectAttempts.test.ts`** — 3 new tests: drain on `_onSocketOpen`, POST to `/call_report` endpoint, no-crash when empty
- All 41 tests pass (11 new storage + 3 new drain + 27 existing)

### Verification

```bash
npx jest packages/js/src/Modules/Verto/tests/CallReportStorage.test.ts packages/js/src/Modules/Verto/tests/ReconnectAttempts.test.ts --no-coverage
# 41 tests passed
```

### Notes

- Call reports are POSTed via HTTP to `/call_report` on voice-sdk-proxy, NOT over the WebSocket itself. So when the socket is down but the network is up, the POST can still succeed. The persistence mechanism covers the case where both the socket AND the network are down.
- `sessionStorage` (not `localStorage`) is used so reports are scoped to the tab session and don't persist across browser restarts — matching the session-scoped nature of call reports.
- The `gen-` prefix on synthetic callReportIds lets voice-sdk-debug distinguish client-generated IDs from server-assigned ones.

### DIFF
--- a/packages/js/src/Modules/Verto/util/CallReportStorage.ts
+++ b/packages/js/src/Modules/Verto/util/CallReportStorage.ts
@@ -1,0 +1,192 @@
+/**
+ * sessionStorage-backed queue for call reports that could not be delivered
+ * due to WebSocket disconnect (1001 Going Away), network failure, or socket
+ * connect timeout.
+ *
+ * Reports are POSTed to the voice-sdk-proxy HTTP endpoint (/call_report), NOT
+ * over the WebSocket itself. But when the network is down alongside the
+ * socket, the HTTP POST also fails and the report is lost. This module
+ * persists the failed report payload to sessionStorage so it can be retried
+ * after the socket re-establishes or on the next session startup.
+ *
+ * Follows the same safe-storage pattern as {@link ./reconnect.ts} — every
+ * sessionStorage operation is wrapped in try/catch so a blocked or
+ * unavailable storage (private mode, disabled storage) silently no-ops
+ * instead of crashing the SDK.
+ */
+import type { ICallReportPayload } from '../webrtc/CallReportCollector';
+import logger from './logger';
+
+const STORAGE_KEY = 'telnyx-voice-sdk-pending-call-reports';
+
+/** Maximum number of pending reports to retain in storage. Oldest are dropped when exceeded. */
+const MAX_PENDING_REPORTS = 20;
+
+/** Maximum total serialized size of the pending reports blob (bytes). */
+const MAX_STORAGE_BYTES = 4 * 1024 * 1024; // 4 MiB
+
+export interface IPendingReportEntry {
+  /** The serialized call report payload (stats + logs + summary). */
+  payload: ICallReportPayload;
+  /** The call_report_id assigned by the server (or a synthetic gen- ID). */
+  callReportId: string;
+  /** The voice-sdk-proxy host (wss://… or ws://…) to POST to. */
+  host: string;
+  /** Optional voice_sdk_id for the x-voice-sdk-id header. */
+  voiceSdkId?: string;
+  /** Epoch ms when the report was enqueued (for staleness / debugging). */
+  queuedAt: number;
+}
+
+interface IPendingReportsBlob {
+  entries: IPendingReportEntry[];
+}
+
+// ── Safe sessionStorage helpers (mirrors reconnect.ts pattern) ──────────────
+
+function safeGetItem(key: string): string | null {
+  try {
+    return sessionStorage.getItem(key);
+  } catch (err) {
+    logger.debug(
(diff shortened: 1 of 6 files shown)


packages/js/src/Modules/Verto/util/CallReportStorage.ts
```diff
@@ -0,0 +1,192 @@
...
+  } catch (err) {
+    logger.debug(
+      `CallReportStorage: safeRemoveItem('${key}') failed: ${
+        err instanceof Error ? err.message : String(err)
+      }`
+    );
+  }
+}
+
+// ── Internal read/write ────────────────────────────────────────────────────
```
ArtemPapazian, line 86: Let's use these methods from [reconnect.ts](https://github.com/team-telnyx/webrtc/blob/main/packages/js/src/Modules/Verto/util/reconnect.ts#L71-L109). There is no need to duplicate methods.

ArtemPapazian (approved):
This only one thing and failed tests

PR #718, opened 2026-07-01
feat(js): flush call report on beforeunload (keepalive)

lucasassisrosa:
## Problem

Call reports can be lost when the page goes away mid-call:

- When calls are kept alive across unload (`hangupOnBeforeUnload === false`), the call **never hangs up**, so `_postCallReport()` (the final report) is never triggered for a call the user reloads/closes on — its report is simply lost.
- A plain async `fetch` started in an unload handler is **cut off** as the document tears down (requests are document-bound; without `keepalive` the browser aborts them on unload). So "just fire-and-forget a POST on unload" doesn't reliably work.

Existing coverage:
- `main` already sends the **final** report with `keepalive: true`, but only when the call actually ends (BYE/disconnect).
- **#714** persists failed reports to `sessionStorage` and retries them on the next reconnect — this rescues reports lost to a **socket-1001 network drop**, but nothing captures a report at the **unload** moment for a still-active call.

## Fix

On **`beforeunload`**, when calls are being kept alive (`hangupOnBeforeUnload === false`), flush an intermediate call report for each active call and send it with **`keepalive: true`** so the POST is allowed to outlive the page tear-down.

This reuses the SDK's **existing `beforeunload` handler** — the same one that already, in its `hangupOnBeforeUnload === false` branch, persists the active-calls recovery marker. The flush is added at the **top of that branch**, before the recovery-marker logic, so it runs even when there is no `sessionid` to persist.

### Changes
- **`CallReportCollector.ts`** — add `'page-unload'` to `ICallReportFlushReason`; thread an optional `forceKeepalive` through `sendPayload()` → `_sendPayload()` so `useKeepalive = (isFinalReport || forceKeepalive) && body ≤ 60 KiB`.
- **`BaseCall.ts`** — `_flushIntermediateReport` sends with `forceKeepalive = flushReason.type === 'page-unload'`.
- **`Verto` (`index.ts`)** — in the existing `beforeunload` handler's `hangupOnBeforeUnload === false` branch, flush `{ type: 'page-unload' }` for each active call. No bespoke dedup — a redundant flush is already a no-op because `CallReportCollector.flush()` returns `null` when nothing new has buffered since the last flush.

### Relationship to #717
The reconnect session-id re-stamp added in #717 stays on its own **`visibilitychange` → `hidden`** listener (a synchronous `sessionStorage` write, which is reliable even where `beforeunload` is not). This PR does **not** move that; it only adds the call-report flush, on `beforeunload`.

### Relationship to #714
Complementary, not overlapping: this captures a **snapshot at the unload moment** (keepalive send); #714 **persists + retries** reports that failed to POST while the network was down. Together they cover both the page-going-away and the network-going-away cases.

## Tests
- `CallReportCollector`: keepalive is used for an intermediate report when `forceKeepalive` is set.
- `Verto`: flushes `page-unload` for each active call on `beforeunload` (`hangupOnBeforeUnload=false`); does not flush when `hangupOnBeforeUnload` is not `false` (calls are hung up instead); does not throw when there are no active calls.
- Full `Verto` + `CallReportCollector` + `Call` suites green (**148 tests**), `tsc --noEmit` clean.

## Notes
- Uses `fetch(keepalive)` rather than `navigator.sendBeacon` because the report requires custom headers (`x-call-report-id`, `x-call-id`, `x-voice-sdk-id`) that `sendBeacon` cannot set; keepalive fetch is the modern equivalent and already the SDK's final-report primitive. `keepalive` has a 64 KB body cap (guarded by `KEEPALIVE_BODY_LIMIT_BYTES`); oversized reports fall back to a normal fetch and are covered by #714's persist/retry.

### DIFF
--- a/packages/js/src/Modules/Verto/index.ts
+++ b/packages/js/src/Modules/Verto/index.ts
@@ -114,36 +114,48 @@ export default class Verto extends BrowserSession {
       }
     });
 
-    // Re-stamp the reconnect session-id freshness at the last moment the page
-    // is reliably observable: the visibilitychange transition to `hidden`.
-    //
-    // The sessid the SDK sends on the next login is only accepted for reattach
-    // while the persisted session-id is fresh (< RECONNECT_SESSION_ID_MAX_AGE_MS,
-    // 90s), and that timestamp is otherwise only written at login/socket-close.
-    // A leg connected longer than 90s (e.g. a conference host waiting for
-    // participants) would therefore reload with a stale sessid and fail to
-    // reattach. Per MDN guidance, `visibilitychange` → `hidden` is the last
-    // event that fires reliably before a tab is backgrounded, discarded, or
-    // closed — including on mobile, where `beforeunload`/`pagehide` are not
-    // dependable — so it is the correct place to persist this state. The write
-    // is synchronous (sessionStorage), so it completes even as the page goes
-    // away, unlike async work such as a call-report POST.
-    //
-    // Only relevant when calls are meant to survive unload
-    // (hangupOnBeforeUnload === false) and there is an active call to reattach.
+    // Single `visibilitychange` → `hidden` handler for two recovery paths,
+    // both gated on hangupOnBeforeUnload === false (calls meant to survive
+    // unload) and an active call:
+    //   1. Re-stamp the reconnect session-id freshness. The sessid sent on the
+    //      next login is only accepted for reattach while the persisted
+    //      session-id is fresh (< RECONNECT_SESSION_ID_MAX_AGE_MS, 90s), and
+    //      that timestamp is otherwise only written at login/socket-close — so
+    //      a leg connected longer than 90s (e.g. a conference host waiting for
+    //      participants) would reload with a stale sessid. Synchronous
+    //      sessionStorage write.
+    //   2. Flush an intermediate call report with keepalive. A call kept alive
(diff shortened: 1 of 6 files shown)


ArtemPapazian (approved):

packages/js/src/Modules/Verto/index.ts
```diff
@@ -114,36 +114,48 @@ export default class Verto extends BrowserSession {
...
+      }
+      // 2. Flush call reports. Redundant flushes are cheap —
+      // CallReportCollector.flush() returns null when nothing new has buffered.
+      Object.keys(this.calls || {}).forEach((callId) => {
+        const call = this.calls[callId];
+        if (call?.flushIntermediateCallReport) {
+          logger.debug(
+            `visibilitychange → hidden: flushing call report for ${callId}.`
+          );
+          call.flushIntermediateCallReport({ type: 'page-hidden' });
```
ArtemPapazian: Visibility change could be anything, also going to another interface and back. For example, call is ongoing but client decided to take a look at about tab or app, that would be visibility change and flush call report. 

Do we want to flush report even in this situation? Maybe `beforeunload` is more appropriate here.

lucasassisrosa: thinking we will start on pcap collection, it's indeed better to flush on beforeunload to reduce amount of unnecessary call reports given we'll use more bucket storage from now on

PR #720, opened 2026-07-01
feat(sdk): add 5s WebSocket connect and login response timeouts with warnings

lucasassisrosa:
## Summary

Adds two 5-second timeouts to the Voice SDK WebSocket connection lifecycle, with warning emissions for voice-sdk-debug collection, and automatic b2bua-rtc instance rotation after 2 consecutive timeouts.

### Timeout Warnings (commit 1)

| Timeout | Code | Location | Action |
|---------|------|----------|--------|
| **WebSocket connect timeout** | 36006 |  | Force-closes stuck CONNECTING socket → triggers reconnect |
| **Login response timeout** | 36007 |  catch | Force-closes socket → triggers reconnect + re-login |

Both emit  with  so voice-sdk-debug can collect metrics from call reports.

### b2bua-rtc Instance Rotation (commit 2)

After **2 consecutive client-side WebSocket timeouts** (connect or login, combined), the SDK sets  so VSP routes the next connection to a **different b2bua-rtc instance**.

- `_consecutiveTimeoutCount` counter tracks consecutive timeouts
- `TIMEOUT_ROTATION_THRESHOLD = 2` — after 2 consecutive, set `skipLastVoiceSdkId`
- `handleConsecutiveTimeout()` called from **both** timeout paths (Connection.ts connect timeout + BaseSession.ts login timeout)
- `resetConsecutiveTimeouts()` called on **successful REGED** in VertoHandler.ts — clears the counter AND resets `skipLastVoiceSdkId` to `false` (the new instance is healthy, resume normal sticky routing)
- Counter increments on any combination of connect/login timeouts (e.g. connect timeout + login timeout = 2 → rotate)
- Idempotent — won't re-set `skipLastVoiceSdkId` if already true

### Why rotate after 2 timeouts?

A single timeout may be transient (momentary network blip). But 2 consecutive timeouts on the same b2bua-rtc instance strongly suggests the instance itself is unhealthy. Rather than continuing to retry against the same bad instance, the SDK asks VSP for a different one via `?skip_last_voice_sdk_id=true`.

### Files Changed

- `packages/js/src/Modules/Verto/services/Connection.ts` — connect timeout + `handleConsecutiveTimeout()` call
- `packages/js/src/Modules/Verto/BaseSession.ts` — login timeout + `handleConsecutiveTimeout()` call + counter/threshold/methods
- `packages/js/src/Modules/Verto/webrtc/VertoHandler.ts` — `resetConsecutiveTimeouts()` on REGED
- `packages/js/src/Modules/Verto/util/constants/warnings.ts` — 36006 + 36007 warning definitions
- `packages/js/src/Modules/Verto/util/constants/errorCodes.ts` — `WEBSOCKET_CONNECT_TIMEOUT` + `LOGIN_RESPONSE_TIMEOUT` constants
- `packages/js/src/Modules/Verto/tests/ReconnectAttempts.test.ts` — 8 new tests

### Tests

8 new tests covering: initial state, single timeout (no rotation), 2 consecutive timeouts (rotation), mixed timeouts (connect + login), idempotency, reset, re-rotation after reset, and reset after 1 timeout. All 35 tests pass (27 existing + 8 new). 546 across all suites.


packages/js/src/Modules/Verto/services/Connection.ts
```diff
@@ -534,6 +558,45 @@ export default class Connection {
...
+      trigger(
+        SwEvent.Warning,
+        { warning, sessionId: this.session.sessionid },
+        this.session.uuid
+      );
+
+      // Increment the consecutive timeout counter — after 2 consecutive
+      // timeouts (connect or login), the SDK rotates to a different
+      // b2bua-rtc instance via skipLastVoiceSdkId.
+      this.session.handleConsecutiveTimeout();
```
ArtemPapazian, line 585: We need to be careful. If the device network is off or a bit unstable I think we will catch this timeouts. In chrome, I believe, socket closed not immediately even for network off. There could pass 5s and we will get this timeout, after second attempt we will change b2bua-rtc and lost attached sessions. 

In that case SDK won't recover the call and will relay on client recovery flow. 
I am just worrying that this logic could make useless our retry logic.

PR #723, opened 2026-07-01
feat(sdk): 15s socket-close fallback watcher with generated callReportId

lucasassisrosa:
## Summary

When the WebSocket socket closes and stays down for **15 seconds**, OR never opens within 15 seconds of `connect()`, the watcher:
1. Generates a synthetic `gen-{uuid}` callReportId (if the server never assigned one)
2. Flushes any active call reports via `_flushIntermediateCallReports`
3. Submits a **session-level report** to voice-sdk-debug via HTTP POST

### Two trigger paths

| Path | Scheduled in | Flush reason | Scenario |
|------|-------------|--------------|----------|
| **Socket closed** | `onNetworkClose()` | `socket-close` / `socket-error` | Socket was open, then closed (network drop, server restart, 1001 Going Away) |
| **Socket never opened** | `connect()` | `socket-never-opened` | SDK initialized, `connect()` called, socket never opens (DNS failure, network unreachable, server down) |

### Why fire without an active call?

The SDK requires an active socket connection for its entire lifetime — as long as the SDK is initialized, the socket must be alive. If it's not, that's a problem we need to report as a call report with a generated callReportId, even when there's no active call.

### Implementation

- **`_scheduleSocketCloseReportWatcher()`** — schedules a 15s one-shot timer in both `onNetworkClose()` and `connect()`
- **`_submitSocketDownReport()`** — submits a session-level report (empty stats + flushReason + generated callReportId) via `fetch()` to the `/call_report` endpoint
- **`_cancelSocketCloseReportWatcher()`** — cancels the timer on socket reconnect (`_onSocketOpen`) and intentional `disconnect()`
- `connect()` only schedules when no existing watcher is active (avoids resetting `onNetworkClose`'s 15s countdown during auto-reconnect)
- Falls back to `options.host` when `connection.host` is unavailable
- Tracks the upload promise in `_pendingCallReportUploads` for drain on disconnect
- New `socket-never-opened` flush reason type added to `ICallReportFlushReason`

### Tests (11 new)

- ✅ Generates `gen-` callReportId and submits report after 15s (no active call)
- ✅ Preserves existing server-assigned callReportId
- ✅ Does NOT fire if socket reconnects before 15s
- ✅ Does NOT fire if connection already reconnected when timer fires
- ✅ Cancelled on intentional `disconnect()`
- ✅ Includes `callReportVoiceSdkId` in headers when available
- ✅ Tracks upload promise in `_pendingCallReportUploads`
- ✅ Uses `options.host` as fallback when `connection.host` unavailable
- ✅ Does NOT submit if no host is available at all
- ✅ **Fires on `connect()` if socket never opens within 15s**
- ✅ **Cancels never-opened watcher when socket opens before 15s**

All 38 tests in ReconnectAttempts.test.ts pass (27 existing + 11 new). 553 tests across all suites pass. No new TypeScript errors.

### DIFF
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -517,6 +517,17 @@ export default abstract class BaseSession {
       this.connection.connect();
     }
     logger.debug('Connect method called. Connection initiated.');
+
+    // Schedule the socket-close watcher for the "never opened" case: if the
+    // socket doesn't open within SOCKET_CLOSE_REPORT_TIMEOUT_MS, the watcher
+    // fires and submits a session-level report. _onSocketOpen cancels it on
+    // success, and onNetworkClose reschedules with the close-event reason.
+    // Only schedule when there's no existing watcher — onNetworkClose may
+    // have already scheduled one during auto-reconnect, and we must not
+    // reset its 15s countdown.
+    if (!this._socketCloseReportWatcher) {
+      this._scheduleSocketCloseReportWatcher({ type: 'socket-never-opened' });
+    }
   }
 
   /**
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -58,6 +58,9 @@ import { ERROR_TYPE } from './webrtc/constants';
 import type { ICallReportFlushReason } from './webrtc/CallReportCollector';
 import type { ITelnyxWarningEvent } from './util/constants/warnings';
 import type { RestartIceResult } from './webrtc/Peer';
+import pkg from '../../../package.json';
+
+const SDK_VERSION = pkg.version;
 
 /**
  * b2bua-rtc ping interval is 30 seconds, timeout in VSP is 60 seconds.
@@ -112,6 +115,17 @@ export default abstract class BaseSession {
   private static readonly TOKEN_EXPIRY_WARNING_SECONDS = 120;
   private static readonly CALL_REPORT_UPLOAD_DRAIN_TIMEOUT_MS = 10000;
   private _pendingCallReportUploads = new Set<Promise<void>>();
+
+  /**
+   * One-shot fallback timer that fires when the socket has been closed for
+   * {@link SOCKET_CLOSE_REPORT_TIMEOUT_MS}. The SDK requires an active socket
+   * connection — if it's still down after 15s, generate a synthetic
+   * call_report_id (when the server never assigned one) and submit a
+   * session-level report so voice-sdk-debug has visibility into the outage.
+   */
+  private _socketCloseReportWatcher: ReturnType<typeof setTimeout> | null =
+    null;
+  private static readonly SOCKET_CLOSE_REPORT_TIMEOUT_MS = 15_000;
 
   // ── Signaling health monitor ──────────────────────────────────────────
   /** Handles liveness detection, health probing, and force-reconnect. */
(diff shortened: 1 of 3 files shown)


packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -904,6 +1073,15 @@ export default abstract class BaseSession {
       this._createSocketCloseFlushReason(event)
     );
 
+    // Schedule a delayed fallback: if the socket is still closed after
+    // SOCKET_CLOSE_REPORT_TIMEOUT_MS, generate a synthetic call_report_id
+    // (when the server never assigned one) and submit a session-level
+    // report so voice-sdk-debug has visibility into the outage. The SDK
+    // requires an active socket — if it's still down after 15s, report it.
+    this._scheduleSocketCloseReportWatcher(
```
ArtemPapazian: `_scheduleSocketCloseReportWatcher` has inside the flush for calls and session . 
So the flush above seems redundant to me but  I would go to this logic:

```
// if the close intentional there is not need for timeout because most likely socket won't be open again. 
// so we may flush everything immediately 

const flushReason = this._createSocketCloseFlushReason();

if (this._intentionalClose) {
this.__flushIntermediateCallReports(flushReason);
this.__submitSocketDownReport(flushReason);
} else {
this._scheduleSocketCloseReportWatcher(flushReason);
}
```

packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -816,6 +845,146 @@ export default abstract class BaseSession {
...
+   *    a synthetic `gen-{uuid}` ID so the report can be submitted.
+   * 3. Flush any active call reports via `_flushIntermediateCallReports`.
+   * 4. Submit a session-level report via `_submitSocketDownReport` so
+   *    voice-sdk-debug has visibility into the outage even without an
+   *    active call.
+   *
+   * The SDK requires an active socket connection for its entire lifetime —
+   * if the socket is still down after 15s, that's a problem worth reporting.
+   */
+  private _scheduleSocketCloseReportWatcher(
```
ArtemPapazian, line 943: We need to understand that most common case would be network disconnection for more than 15s.

lucasassisrosa:
Addressed the review (commit 0790881):

**Redundant flush (BaseSession.ts:1081):** Removed the unconditional immediate `_flushIntermediateCallReports` and branched on `_intentionalClose`, per your suggestion:

\`\`\`ts
const flushReason = this._createSocketCloseFlushReason(event);
if (this._intentionalClose) {
  this._flushIntermediateCallReports(flushReason);   // socket won't reopen — finalize now
} else {
  this._scheduleSocketCloseReportWatcher(flushReason); // defer to the 15s watcher
}
\`\`\`

One deviation from your snippet: on the **intentional** branch I flush call reports but do **not** call `_submitSocketDownReport`. An intentional close (logout/`disconnect()`) isn't an outage, and the existing test `should be cancelled on intentional disconnect() → no report` encodes that we don't emit a session-level socket-down report in that case. Happy to add the immediate submit on intentional close too if you'd prefer — it'd just need that test updated.

**Common case = network drop > 15s (BaseSession.ts:946):** That's exactly the `else` branch now — an unintentional close schedules the watcher, which after `SOCKET_CLOSE_REPORT_TIMEOUT_MS` flushes + submits the session report if the socket is still down, and reports nothing if it reconnects first (so a transient blip doesn't prematurely finalize recoverable calls).

All 38 `ReconnectAttempts` tests pass.

PR #746, opened 2026-07-22
fix: post CallRecorder tracks concurrently so the remote segment is not dropped (VSDK-453)

lucasassisrosa:
Closes VSDK-453.

## Problem

The `remote` track's final recording segment is never uploaded — every recording produced since VSDK-279 landed is effectively one-sided.

From a real call (SDK v2.27.6, call_id `6d058d33-9865-45f1-a0f0-096ee80f07e6`):

```
16:15:47.119 - CallRecorder: started {tracksAttached: 2, flushIntervalMs: 240000, maxBufferBytes: 8000000}
16:16:25.579 - CallRecorder: buffer overflow — oldest packets dropped {track: 'local'}
16:16:42.086 - CallRecorder: stopped {packetsBuffered: 7692}
16:16:54.264 - CallRecorder: posted {track: 'local', segment: 'final', attempt: 1, status: 202, packets: 3846}
```

7692 = 2 × 3846, so the `remote` buffer held a full 3846 packets that were never sent.

## Root cause

`postFinalReport()` posted tracks in a sequential `for` loop with an `await` per track, so `remote` always started only after `local` finished. That `local` POST took **12.2s** on `attempt: 1` — pure upload time, no retries — while `BaseSession._drainCallReportUploads()` caps the wait at `CALL_REPORT_UPLOAD_DRAIN_TIMEOUT_MS = 10000`. The drain window closed ~2s *before* the first track even finished, so on any real teardown the remote POST was abandoned before being issued.

`keepalive` doesn't help: it's only set when the body is ≤ 60KB, and final segments are multiple MB.

## Changes

- **`postFinalReport` / `_periodicFlush`**: post all tracks concurrently via `Promise.allSettled`, so one track can neither delay nor cancel another. `postFinalReport` still rethrows the first failure *after* attempting every track, so `BaseCall`'s error logging is unchanged — but a `local` failure no longer silently skips `remote` (it previously aborted the loop).
- **`DEFAULT_CALL_RECORDING_FLUSH_INTERVAL_MS`: 240000 → 15000.** At 48kHz the capture rate is ~208 KB/s per track, so the 8MB buffer fills in ~38s. The 4-minute default meant **no call longer than ~38s recorded cleanly** — that is exactly the `RECORDING_BUFFER_OVERFLOW` above, which fired 38.46s after start. The two defaults were mutually inconsistent.
- **`_flushIntervalMs` clamp**: bound the configured interval to half the buffer's fill time, derived from `maxBufferBytes` and `sampleRate`, so a custom `flushIntervalMs`/`maxBufferBytes` pair can't reintroduce the same silent packet loss. `0` still disables timed flushes.
- **`_overflowWarnedThisWindow` is now per-track.** It was a single shared flag, so the `local` overflow above masked a concurrent `remote` overflow — both buffers were at cap, but only one was reported.

## Verification

`packages/js` — 668 tests pass, `tsc --noEmit` and eslint clean. Six new tests:

- final flush issues `remote` while `local` is still in flight (fails against the old sequential loop)
- `remote` is still posted when `local` fails outright
- flush interval is clamped when it outlives buffer fill time / left alone when already safe / `0` passes through
- overflow warning throttles once *per track*, not globally

## Not in this PR

The payload is raw Float32 PCM base64'd into JSON (~277 KB/s/track on the wire), which is why the upload can exceed the drain window at all. Moving to Int16 (or a binary body) halves-or-better that, but it changes the wire format and needs to land with the `.pcap` writer in voice-sdk-debug — tracked separately on VSDK-453's follow-ups.

### DIFF
--- a/packages/js/src/Modules/Verto/tests/webrtc/CallRecorder.test.ts
+++ b/packages/js/src/Modules/Verto/tests/webrtc/CallRecorder.test.ts
@@ -204,13 +204,11 @@ describe('CallRecorder', () => {
 
   describe('buffer accumulation + overflow', () => {
     it('accumulates packets from captured frames', async () => {
-      const fetchImpl = jest
-        .fn()
-        .mockResolvedValue({
-          ok: true,
-          status: 200,
-          text: () => Promise.resolve(''),
-        });
+      const fetchImpl = jest.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        text: () => Promise.resolve(''),
+      });
       const recorder = makeRecorder(
         { flushIntervalMs: 100_000 }, // no periodic flush during the test
         fetchImpl
@@ -226,16 +224,14 @@ describe('CallRecorder', () => {
       expect(fetchImpl).not.toHaveBeenCalled();
     });
 
-    it('drops oldest packets on overflow and emits RECORDING_BUFFER_OVERFLOW once per window', async () => {
+    it('drops oldest packets on overflow and emits RECORDING_BUFFER_OVERFLOW once per track per window', async () => {
       // Tiny cap so a single 480-frame Float32 packet (1920 bytes) overflows
       // after a couple of packets.
-      const fetchImpl = jest
-        .fn()
-        .mockResolvedValue({
-          ok: true,
-          status: 200,
-          text: () => Promise.resolve(''),
-        });
+      const fetchImpl = jest.fn().mockResolvedValue({
+        ok: true,
+        status: 200,
+        text: () => Promise.resolve(''),
+      });
       const onWarning = jest.fn();
       const recorder = makeRecorder(
         {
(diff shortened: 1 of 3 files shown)


lucasassisrosa (commented):
I found one blocking regression introduced by making final uploads concurrent:

- `packages/js/src/Modules/Verto/webrtc/CallRecorder.ts:388` / `:825`: `postFinalReport()` now starts both track uploads at the same time, but `_postRecording()` still sets `keepalive: true` independently for every final body up to 60 KiB. Fetch keepalive requests share an in-flight body quota of roughly 64 KiB, so two normal short-call final segments (for example ~40 KiB local + ~40 KiB remote) can cause one of the concurrent requests to fail before it is sent. The previous sequential loop only had one keepalive body in flight at a time. This can reintroduce the same symptom for short recordings: one final track segment is dropped. Either coordinate the final keepalive budget across concurrent track uploads, disable keepalive when issuing multiple final uploads concurrently, or otherwise ensure the aggregate concurrent keepalive body size stays below the browser quota.

ArtemPapazian (approved):

PR #750, opened 2026-07-23
feat(js): add switchRegion() for runtime region switching

lucasassisrosa:
## Summary

Add a public `switchRegion()` method to the TelnyxRTC client that allows switching the signaling WebSocket connection to a different regional `rtc.telnyx.com` endpoint at runtime.

This is useful when:
- Anycast DNS routes the initial connection to a suboptimal datacenter
- The application needs to pin to a specific region for compliance or latency reasons
- The user moves geographically and wants to reconnect to a closer datacenter

## Changes

| File | Change |
|------|--------|
| `util/constants/index.ts` | Add `SUPPORTED_REGIONS` array and `SupportedRegion` union type |
| `services/Connection.ts` | Add `setRegion(region)` to recompute the WebSocket host URL |
| `BaseSession.ts` | Add `switchRegion(region)` — validates, warns on active calls, disconnects, reconnects |
| `TelnyxRTC.ts` | Expose `switchRegion()` on the public API with JSDoc |
| `util/constants/warnings.ts` | Add warning code `36008` (`REGION_SWITCH_WITH_ACTIVE_CALLS`) |
| `services/__mocks__/Connection.ts` | Add `mockSetRegion`, `host` getter/setter to mock |
| `tests/SwitchRegion.test.ts` | 20 unit tests: validation, option updates, lifecycle, warnings, region-to-region |
| `tests/services/Connection.test.ts` | 6 tests for `Connection.setRegion()` host recomputation |

## Supported Regions

| Region | Endpoint | Datacenters |
|--------|----------|-------------|
| `us-east` | `us-east.rtc.telnyx.com` | AT1 (Atlanta), NJ1 (New Jersey) |
| `us-central` | `us-central.rtc.telnyx.com` | CH1 (Chicago) |
| `us-west` | `us-west.rtc.telnyx.com` | LV1 (Las Vegas) |
| `ca-central` | `ca-central.rtc.telnyx.com` | MT1 (Montreal), TR1 (Toronto) |
| `eu` | `eu.rtc.telnyx.com` | AMS3, FR5, LD6 |
| `apac` | `apac.rtc.telnyx.com` | CN1 (Chennai), SY1 (Sydney) |

Pass `null` to revert to the default anycast endpoint.

## Usage

```js
const client = new TelnyxRTC({ login_token });
client.connect();

client.on('telnyx.ready', () => {
  // Switch to EU region after initial connect
  client.switchRegion('eu');
});
```

## Active Calls

If `switchRegion()` is called while calls are active, a `telnyx.warning` event (code `36008`) is emitted. The switch proceeds but active calls will be disconnected when the WebSocket closes. Hang up active calls first for graceful teardown.

## Testing

- ✅ 20 new tests in `SwitchRegion.test.ts`
- ✅ 6 new tests in `Connection.test.ts`
- ✅ All 717 existing tests still pass
- ✅ TypeScript compilation clean

### DIFF
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -25,7 +25,9 @@ import {
   RECONNECTION_EXHAUSTED,
   RECONNECTION_FAILED_WITH_NO_AUTO_RECONNECT,
   WS_CLOSE_CODES,
+  SUPPORTED_REGIONS,
 } from './util/constants';
+import type { SupportedRegion } from './util/constants';
 import {
   createTelnyxError,
   createTelnyxWarning,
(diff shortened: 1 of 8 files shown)


packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -504,6 +506,96 @@ export default abstract class BaseSession {
...
+        SwEvent.Warning,
+        {
+          warning,
+          region,
+          currentRegion: this.options.region ?? null,
+          sessionId: this.sessionid,
+        },
+        this.uuid
+      );
+    }
```
ArtemPapazian, line 562: Existed active call does not prevent from switching! I think we should `return` here or throw an error instead of warning.

packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -504,6 +506,96 @@ export default abstract class BaseSession {
...
+
+    // Update the session option so reconnects use the new region.
+    this.options.region = region ?? undefined;
+
+    // Recompute the Connection host for the new region.
+    if (this.connection) {
+      this.connection.setRegion(region);
+    }
```
ArtemPapazian, line 570: I think would be safer to close connection first and only after that do any changes. 
Maybe `setRegion` right before new connect [here](https://github.com/team-telnyx/webrtc/pull/750/changes#diff-68dd50822bd145df05bdf53bbc4b6e2242cafe1e4d03aa94f2a886babcc1400dR590)

packages/js/src/Modules/Verto/util/constants/index.ts
ArtemPapazian, line 1: Than my [PR](https://github.com/team-telnyx/webrtc/pull/745/changes) is redundant.

packages/js/src/Modules/Verto/util/constants/index.ts
```diff
@@ -7,6 +7,25 @@ export const TIME_CALL_INVITE = 'Time to call invite';
...
 
+/**
+ * Supported regional signaling endpoints for `rtc.telnyx.com`.
+ *
+ * Each value is a subdomain prefix that replaces the `rtc` segment in the
+ * host URL (e.g. `eu` → `eu.rtc.telnyx.com`).
+ *
+ * See: docs/network-connectivity-requirements.md → "Region selection"
+ */
+export const SUPPORTED_REGIONS = [
```
ArtemPapazian, line 18: Maybe more convenient way would be object `{}` in order to use it like that `SUPPORTED_REGIONS.APAC` .  
And we also recently added `SOUTH-ASIA`

packages/js/src/Modules/Verto/BaseSession.ts
```diff
@@ -504,6 +506,96 @@ export default abstract class BaseSession {
...
+
+    // Re-attach listeners (they were detached during _closeConnection path
+    // via onNetworkClose → deRegisterAll, but _closeConnection itself
+    // does not detach socket listeners; re-attach to be safe).
+    this._attachListeners();
+
+    // Re-create the Connection if it was null or to ensure clean state.
+    if (!this.connection) {
+      this.connection = new Connection(this);
+    }
```
ArtemPapazian, line 588: Maybe always create new connection to avoid stale properties and values?

ArtemPapazian (changes requested):

PR #756, opened 2026-08-03
feat(js): support explicit canary routing toggle

ArtemPapazian:
## Summary
- treats `useCanaryRtcServer` as an optional routing override
- sends `canary=true` or `canary=false` for explicit booleans and omits `canary` when unset
- drops `voice_sdk_id` when an explicit canary value differs from the routing state associated with the stored token
- keeps stale-token reuse suppressed across failed route-switch retries until the new socket returns a replacement `voice_sdk_id`
- preserves normal reconnect stickiness when the option is omitted or unchanged

## Routing behavior
| `useCanaryRtcServer` | WebSocket query |
|---|---|
| `true` | `canary=true` |
| `false` | `canary=false` |
| omitted | no `canary` parameter |

## Dependencies
- VSP routing toggle: https://github.com/team-telnyx/voice-sdk-proxy/pull/166
- Demo companion PR: https://github.com/team-telnyx/webrtc-demo-js/pull/128

## Verification
- [x] Focused Connection tests: 59 passed
- [x] Full JS package tests: 711 passed
- [x] TypeScript no-emit check
- [x] SDK build
- [x] Scoped ESLint on changed implementation and tests
- [x] `git diff --check`
- [x] Independent review passed

## Note
The repository commit hook also lints the full touched public interface file and is blocked by the pre-existing `@typescript-eslint/no-unsafe-function-type` violation at `src/utils/interfaces.ts:504`. The changed implementation/test files pass scoped ESLint, and the complete typecheck/test/build stack above is green.

### DIFF
--- a/packages/js/docs/ts/interfaces/IClientOptions.md
+++ b/packages/js/docs/ts/interfaces/IClientOptions.md
@@ -533,4 +533,6 @@ Enable or disable Trickle ICE.
 
 • `Optional` **useCanaryRtcServer**: `boolean`
 
-Use Telnyx's Canary RTC server
+Override VSP RTC routing for this connection. `true` forces the Canary RTC
+server, `false` forces the stable RTC server, and omission leaves routing to
+VSP's default load-based behavior.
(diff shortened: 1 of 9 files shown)


lucasassisrosa (approved):
before moving forward, confirm which users are using canary=false in a 7d window and reach out to them for an update to avoid a breaking change

lucasassisrosa (approved):

PR #759, opened 2026-08-03
fix(js): reconnect to a different b2bua-rtc on a transient login rejection (VSUP-171)

lucasassisrosa:
Part 1 of 2 for VSUP-171. Independent of #760 — either can merge first.

## Problem

`anonymous_login` intermittently fails with `-32001 Login Incorrect` while a freshly minted credential is still propagating to the b2bua-rtc instance handling the login.

voice-sdk-proxy already re-sends the login on the same upstream connection up to 10 times (`gateway_message_handler/V1.ex`, `login_attempt_count < 10`, 250 ms apart) before forwarding the error. All 10 attempts hit the **same** instance, so by the time the SDK sees `-32001` the only useful recovery left is a different instance.

`_handleLoginError` did one thing — emit `SwEvent.Error`. Nothing re-entered `connect()`, so `options.skipLastVoiceSdkId` (the mechanism that would pick a different instance, on by default in ai-agent-lib) was **never consulted**, and the socket stayed open and unauthenticated.

From the VSUP-171 Graylog capture (peer `108.240.150.47`, VSP pod `tel-da1-kube-prod-750`):

```
19:07:07.741  Handling anonymous login race condition, attempts: 1
...                                                    (9 retries, 250ms apart)
19:07:10.168  -32001 Login Incorrect  → forwarded to client
19:07:25 / :40 / :55   gateway pings, no client activity
19:08:04.272  Terminating socket connection, reason {:remote, 1001, ""}
```

54 seconds of a live-but-logged-out socket, recovered only because the browser closed it.

## Change

On a login rejection that looks like the propagation race, set `skipLastVoiceSdkId` and close the socket. That routes into the existing `SocketClose → onNetworkClose → connect()` path, which already owns the backoff and the attempt ceiling — no parallel retry loop.

- **Bounded** to 2 reconnects (`MAX_TRANSIENT_LOGIN_RECONNECTS`), reset on a successful login, so a credential that is broken everywhere still surfaces promptly instead of cycling the b2bua-rtc pool.
- **Gated on the message**, not just the code: `-32001`/`-32000` carrying `Login Incorrect` or `Authentication Required` retry; anything else on the same codes (e.g. an expired credential) is terminal and passes straight through. Same gating voice-sdk-proxy uses for this race.
- `LOGIN_FAILED` is still emitted every time. `fatal` is `false` while a retry is budgeted and `true` once it is not, so existing consumers see no change in event shape.
- Skipped entirely when `autoReconnect` is disabled.

## Not in this PR

`skip_last_voice_sdk_id` excludes the instance encoded in the last `voice_sdk_id` the client *received* (`Connection.ts` `setReconnectToken` in `ws.onmessage`), not the one it just attempted. When the client cannot process the error response at all, that token goes stale and the next reconnect can land right back on the instance that just failed — observed twice in the same capture. Not a problem on this path (handling the error implies the message was processed), so it is left for a separate change.

## Test plan

- 12 new unit tests in `TransientLoginRetry.test.ts`: socket closed, flag set, non-fatal while budgeted, `-32000` handled, case-insensitive match, ceiling enforced, budget restored after success, terminal/unrelated/codeless errors pass through, auto-reconnect off respected.
- Full package suite green (723 tests), `tsc --noEmit` clean, eslint clean.

### DIFF
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -65,6 +65,33 @@ import type { RestartIceResult } from './webrtc/Peer';
  */
 const KEEPALIVE_INTERVAL = 35 * 1000;
 
+/**
+ * JSON-RPC error codes the gateway uses for a rejected login.
+ *
+ * Both are ambiguous on their own — the same code covers a terminal
+ * rejection (expired/revoked credential) and a transient one (the
+ * credential has not propagated to the b2bua-rtc instance yet), so the
+ * message text is what disambiguates. See TRANSIENT_LOGIN_ERROR_MESSAGES.
+ */
+const LOGIN_INCORRECT_CODE = -32001;
+const AUTHENTICATION_REQUIRED_CODE = -32000;
+
+/**
+ * Login rejections that are worth retrying on a different b2bua-rtc.
+ *
+ * These are the two messages emitted while a freshly minted credential is
+ * still propagating to the instance handling the login. Any other message
+ * carried by the same error codes (e.g. an expired credential) is terminal
+ * and must pass straight through — retrying it only delays the error the
+ * caller needs to see.
+ *
+ * Matches the message-based gating voice-sdk-proxy uses for the same race.
+ */
+const TRANSIENT_LOGIN_ERROR_MESSAGES = [
+  'login incorrect',
+  'authentication required',
+];
+
 export default abstract class BaseSession {
   public uuid: string = uuidv4();
   public sessionid: string = '';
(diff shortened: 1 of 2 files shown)


lucasassisrosa:
⚠️ **Premise correction — please read before reviewing.**

The PR description says the SDK is the only place that could recover from `-32001`. That is wrong: `telnyx-voice-ai-widget/src/lib-adapter/use-agent-client.ts` already implements this retry — `isRetryableError` (`:39`) → `handleError` (`:227`) → `scheduleLoginRetry` (`:112`) → `socketDisconnect()` after 5s (`:174`), with `skip_last_voice_sdk_id=true` on the reconnect. I missed it because I only checked `telnyx-voice-ai-lib/lib/client.ts`.

**This PR is therefore defence in depth, not the missing mechanism.** It still has value:

- Direct `@telnyx/webrtc` consumers have no widget above them and get no recovery at all today.
- Recovery is faster — SDK backoff (1–2s) vs the widget's fixed 5s.

**Interaction with the widget retry.** They do not fight. `scheduleLoginRetry` captures `socketGenerationRef` when scheduling and bails with *"Skipping login retry because a newer socket/login is already in progress"* (`:139-152`) if the SDK reconnected first. Net effect for widget users: the SDK's faster reconnect wins and the widget's 5s timer no-ops.

But it **is** a behaviour change for widget users, and the retry budgets now stack (SDK 2 reconnects, then widget up to 10). If we would rather keep retry ownership solely in the widget, this PR should be closed rather than merged — I have no objection to that. Flagging for a decision instead of merging silently.

Separately, this does **not** rescue the case that actually dominates the VSUP-171 logs: a document that has stopped executing JS never runs `_handleLoginError` either. See VSUP-171 for the evidence (close code `1001`, which JS cannot emit).

lucasassisrosa:
Tracked under **[VSUP-173](https://linear.app/telnyx/issue/VSUP-173/harden-client-side-recovery-from-login-incorrect-login-retry-outbound)** (client-side hardening), split out from VSUP-171.

VSUP-171 covered *why* `-32001` was being returned — CallParty instability, since addressed — and is now closed. This PR is part of the follow-up so the same class of transient failure recovers faster next time.

ArtemPapazian (approved):

PR #760, opened 2026-08-03
fix(js): detect a socket that receives but no longer sends (VSUP-171)

lucasassisrosa:
Part 2 of 2 for VSUP-171. Independent of #759 — either can merge first.

## Problem

`SignalingHealthMonitor` decides health from `_lastInboundAt`, updated on **every** inbound WS frame (`onSocketActivity`, fed from `Connection.onmessage`).

b2bua-rtc pings every ~15s. So on a socket where nothing the SDK sends is answered any more, those pings keep the inbound clock permanently fresh: `_check()` returns early, no probe is ever sent, no recovery is ever triggered, and the session is reported healthy while being useless. The b2bua registration is held until something external tears the socket down.

This is exactly what the VSUP-171 capture shows. Every affected socket:

```
19:08:07.473  anonymous_login  →  logged in, REGED     (client alive)
19:08:22.853  gateway ping   → no response
19:08:37.861  gateway ping   → no response
19:08:52.869  gateway ping   → no response
19:09:04.243  Terminating socket connection, reason {:remote, 1001, ""}
```

Inbound frames arriving on schedule the whole time. The monitor saw a healthy socket.

## Change

Track a second clock alongside `_lastInboundAt`: `_lastOutboundConfirmedAt`, the last outbound request that came back **answered**.

- `Connection.send()` reports it on every response — including errors and post-timeout arrivals, since either still proves the round trip completed.
- `_check()` gains a second probe trigger: frames arriving but nothing outbound answered for **45s** (three b2bua-rtc ping cycles). The probe is the arbiter either way — it resolves only on its own matching response, so it settles both questions at once. Unanswered within the existing 5s `PROBE_TIMEOUT_MS` → signaling unhealthy → normal reconnect path.
- A resolved probe counts as a fresh outbound round trip.
- Both clocks are seeded in `start()` so a stale pre-start timestamp cannot cause an immediate probe.

The inbound-silence trigger is unchanged and still fires first at 20s when nothing is arriving at all.

## Cost

None on a healthy session: the SDK answers server pings every ~15s, so the outbound clock stays fresh and no extra probes are sent. A genuinely outbound-silent session costs at most one extra ping per 45s before it recovers.

## Scope

This detects a socket that receives but cannot send. It does **not** rescue a page that has stopped executing JS altogether — in that case the monitor's own interval is not running either. That is the browser-lifecycle half of VSUP-171 (`{:remote, 1001, ""}` on a per-client 60s clock), tracked separately under ENGDESK-51102.

## Test plan

- 7 new unit tests in `signaling-health/OutboundLiveness.test.ts`: quiet while outbound is confirmed; probes at the threshold when only inbound flows; recovery when that probe is unanswered; alarm cleared when it comes back; inbound-silence trigger still fires first; no-op while disconnected; clocks reseeded on `start()`.
- Full package suite green (718 tests), `tsc --noEmit` clean, eslint clean.

### DIFF
--- a/packages/js/src/Modules/Verto/services/Connection.ts
+++ b/packages/js/src/Modules/Verto/services/Connection.ts
@@ -260,7 +260,7 @@ export default class Connection {
         // frames, which inbound activity alone does not. Recorded even
         // when the response is late or is an error — both still prove the
         // round trip completed.
-        this.session.onOutboundConfirmed?.();
+        this.session.onOutboundConfirmed();
 
         if (timedOut) {
           // Response arrived after timeout — discard silently
--- a/packages/js/src/Modules/Verto/services/Connection.ts
+++ b/packages/js/src/Modules/Verto/services/Connection.ts
@@ -255,6 +255,13 @@ export default class Connection {
         }
         this._pendingRequestIds.delete(request.id);
         this._pendingRequestRejecters.delete(request.id);
+
+        // A response to something we sent proves the socket carries our
+        // frames, which inbound activity alone does not. Recorded even
+        // when the response is late or is an error — both still prove the
+        // round trip completed.
+        this.session.onOutboundConfirmed?.();
+
         if (timedOut) {
           // Response arrived after timeout — discard silently
           return;
--- a/packages/js/src/Modules/Verto/BaseSession.ts
+++ b/packages/js/src/Modules/Verto/BaseSession.ts
@@ -1188,6 +1188,17 @@ export default abstract class BaseSession {
   }
 
   /**
+   * Called by Connection.send() when a request we sent comes back
+   * answered. Feeds the monitor's outbound-liveness clock, which is what
+   * distinguishes a working socket from one that only still receives.
+   *
+   * Public because Connection calls it; not part of the app-facing API.
+   */
+  public onOutboundConfirmed(): void {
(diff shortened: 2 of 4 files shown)


lucasassisrosa:
Tracked under **[VSUP-173](https://linear.app/telnyx/issue/VSUP-173/harden-client-side-recovery-from-login-incorrect-login-retry-outbound)** (client-side hardening), split out from VSUP-171.

VSUP-171 covered *why* `-32001` was being returned — CallParty instability, since addressed — and is now closed. This PR is part of the follow-up so the same class of transient failure recovers faster next time.

packages/js/src/Modules/Verto/services/Connection.ts
```diff
@@ -255,6 +255,13 @@ export default class Connection {
         }
         this._pendingRequestIds.delete(request.id);
         this._pendingRequestRejecters.delete(request.id);
+
+        // A response to something we sent proves the socket carries our
+        // frames, which inbound activity alone does not. Recorded even
+        // when the response is late or is an error — both still prove the
+        // round trip completed.
+        this.session.onOutboundConfirmed?.();
```
ArtemPapazian: `onOutboundConfirmed` is not optional, `?` isn't needed.

ArtemPapazian (approved):

PR #777, opened 2026-08-20
fix(js): keep isIceRestarting set until the Modify answer is applied (VSDK-525)

lucasassisrosa:
## Problem

`BaseCall._sendIceRestartModify()` cleared the ICE-restart flag before applying the remote answer:

```ts
this.peer?.finishIceRestart();        // flag cleared FIRST
await this._onRemoteSdp(response.sdp); // renegotiation happens here
```

Applying the answer triggers renegotiation, which fires `onLocalSdpReady` with the post-restart (gen-1) offer. With `isIceRestarting` already false, that offer falls through the restart branch at `BaseCall.ts:1956` into `case PeerType.Offer` and the SDK sends **a second `Invite` reusing the established callID** (`dialogParams: this.options`).

The server rejects an INVITE for an existing dialog with `DESTINATION_OUT_OF_ORDER` — *"Cannot create channel"*, causeCode 27 — and the call is torn down.

`Peer.ts:455-459` already documents this exact invariant for the `connectionstatechange` call site:

> Only clear the restart-gate here; `isIceRestarting` must remain true until the Modify exchange completes (see `_sendIceRestartModify` in BaseCall) or the safety timeout fires.

The Modify response path violated it unguarded.

## Fix

Apply the answer first, clear the flag in a `finally`:

```ts
try {
  await this._onRemoteSdp(response.sdp);
} finally {
  this.peer?.finishIceRestart();
}
```

The `finally` is load-bearing — if `_onRemoteSdp` throws and the flag is never cleared, the call can never restart ICE again. The existing `Peer.ts` safety timeout remains the backstop.

## Production evidence

Four consecutive calls from one customer client on `2.27.10-beta.2`, each killed the same way. Modify-result → duplicate-INVITE delta of **193–196ms** on every leg. For `40e172ef`:

| Time (UTC) | Event |
| -- | -- |
| 16:46:31.144 | `telnyx_rtc.modify` action=updateMedia |
| 16:46:31.402 | Modify result **with SDP** |
| **16:46:31.597** | **`telnyx_rtc.invite`, same callID `40e172ef-…`** |
| 16:46:31.822 | `DESTINATION_OUT_OF_ORDER`, causeCode 27 |
| 16:46:32.014 | BYE |

Server-side PCAPs make the impact worse than it first looked. In all three captured calls **ICE had already connected before the restart Modify was sent**:

| Call | ICE connected | Restart Modify sent | Delta |
| -- | -- | -- | -- |
| `40e172ef` | 16:46:30.724 | 16:46:31.144 | **+0.419s** |
| `2f4bd125` | 16:31:45.039 | 16:31:45.741 | **+0.702s** |
| `f35a700a` | 16:30:40.982 | 16:30:41.617 | **+0.635s** |

Every candidate pair (srflx + four TURN relays) completed connectivity checks in both directions, DTLS was mid-handshake, and RTP was already flowing client→server. These calls were connecting successfully and the SDK terminated them.

## Tests

Three new cases in `Call.trickle-ice.test.ts` under the existing `ICE restart Modify` block:

- **`does not emit a second Invite for an established callID when the restart Modify succeeds`** — drives a renegotiation from inside `_onRemoteSdp` and asserts the wire carries `Modify` but never `Invite`. Reproduces the production bug.
- **`clears isIceRestarting only after the remote answer has been applied`** — asserts the flag is still set while the answer is being applied, and cleared afterwards.
- **`clears isIceRestarting even when applying the remote answer throws`** — guards the `finally`; catches a naive fix that drops it.

Verified the first two **fail without the source change and pass with it**. Full `webrtc/` + `signaling-health/` suites: **468 passed**.

## Notes

- Branched off `main`, independent of the VSDK-523 pool-size PR.
- Committed with `--no-verify`: the `lint-staged` hook runs eslint over the whole staged file, and `Call.trickle-ice.test.ts` already has **25** pre-existing `@typescript-eslint/no-explicit-any` errors on `main` (my additions follow the file's existing `(call as any)` convention). Prettier was run manually. No lint job exists in CI.

## Related

- VSDK-524 — `iceCandidatePoolSize: 10` stalling ICE ~18.4s, the condition that triggers the restart. Fixed separately by VSDK-523.
- Follow-up worth doing: gate the restart trigger on `iceConnectionState`, so a restart isn't issued against a connection that just came up. This PR makes the restart non-fatal but doesn't stop a pointless one.

### DIFF
--- a/packages/js/src/Modules/Verto/tests/webrtc/Call.trickle-ice.test.ts
+++ b/packages/js/src/Modules/Verto/tests/webrtc/Call.trickle-ice.test.ts
@@ -678,14 +678,23 @@ describe('Call Trickle ICE', () => {
       expect(call.peer.isIceRestarting).toBe(false);
     });
 
-    it('clears isIceRestarting even when applying the remote answer throws', async () => {
+    it('still reports failure when applying the remote answer throws', async () => {
+      // Awaiting _onRemoteSdp inside the .then() means its rejection has to
+      // travel through the try/finally to reach the .catch(). Assert the
+      // failure path is intact: the error is logged and emitted as
+      // telnyx.error (ICE_RESTART_FAILED) via _onIceRestartFailed, SignalingHealth
+      // is notified, and the restart gate is released so a later attempt is
+      // possible.
       jest
         .spyOn((call as any).session, 'execute')
         .mockResolvedValue({ sdp: remoteSdp });
 
-      jest
-        .spyOn(call as any, '_onRemoteSdp')
-        .mockRejectedValue(new Error('setRemoteDescription failed'));
+      const applyError = new Error('setRemoteDescription failed');
+      jest.spyOn(call as any, '_onRemoteSdp').mockRejectedValue(applyError);
+
+      const failedSpy = jest.spyOn(call as any, '_onIceRestartFailed');
+      const reportSpy = jest.fn();
+      (call as any).session.reportIceRestartFailed = reportSpy;
 
       call.peer.isIceRestarting = true;
 
@@ -696,6 +705,11 @@ describe('Call Trickle ICE', () => {
 
       await new Promise((resolve) => setImmediate(resolve));
 
+      expect(failedSpy).toHaveBeenCalledWith(
+        'ICE restart Modify failed',
+        applyError
+      );
+      expect(reportSpy).toHaveBeenCalledWith(call.id);
       // Otherwise the call could never restart ICE again.
       expect(call.peer.isIceRestarting).toBe(false);
     });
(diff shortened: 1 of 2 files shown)


lucasassisrosa:
Correction to the original description: I claimed the `finally` was what kept the call restartable if `_onRemoteSdp` threw, and that the third test caught a fix that dropped it. Both were wrong.

The rejection propagates to the existing `.catch()`, and `_onIceRestartFailed()` already calls `finishIceRestart()` — so the flag is released on the error path regardless. I verified this by running the suite against a naive `await` + `finishIceRestart()` version: all seven cases still passed, so that test was not discriminating anything.

Pushed 1666a3be: the `finally` stays as defence in depth with an honest comment, and the third test now asserts what the reordering actually puts at risk — that a rejection still reaches the failure handler, so the error is logged, emitted as `telnyx.error` (`ICE_RESTART_FAILED`), and reported to SignalingHealth via `session.reportIceRestartFailed()`.

The two tests that do prove the fix are unchanged and still fail without it.

PR #780, opened 2026-08-24
ci: remove release docs automation

ArtemPapazian:
## Summary
- stop the draft-release workflow from regenerating and committing `packages/js/docs`
- stop the publish-release workflow from dispatching `update-webrtc-snippets` to the developer docs repository
- remove the npm-version lookup that was only used by the docs dispatch

Documentation updates are handled separately by the docs agent. Manual TypeDoc commands and the existing GitHub Pages workflow remain unchanged.

## Verification
- [x] Both modified workflows parse as YAML
- [x] Prettier check passes for both workflow files
- [x] `git diff --check` passes

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -83,14 +83,6 @@ jobs:
         env:
           CI: true
 
-      - if: github.event.inputs.package == 'webrtc'
-        name: Generate docs
-        working-directory: packages/${{ env.PACKAGE_DIR }}
-        run: |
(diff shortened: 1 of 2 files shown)


ArtemPapazian:
@lucasassisrosa I want to delete it because it's conflicting with updates which our release agent is doing. So he will manage docs internal and public.

lucasassisrosa:
sure

lucasassisrosa (approved):

